PR #4, opened 2021-06-24
Change `const system` declaration to `var system`

### LINKED ISSUE #3: Changing `const system` declaration in `html.py` to `var system`?
Hello from the Princeton office! We are excited about `brax` and have started playing around with it.

One snag in our workflow: the generated JavaScript for `html.py:render` declares `system` as a `const`, so re-evaluating an Jupyter notebook cell calling `html.render` will fail. We've forked and changed the `const` to `var`, but hesitated submitting a PR since there may have been some important reason we missed for choosing `const`.

Anyway, just wanted to check!

danielsuo:
Fixes #3

### DIFF
--- a/brax/io/html.py
+++ b/brax/io/html.py
@@ -74,7 +74,7 @@ body { margin: 0; }
   </head>
 
   <script type="application/javascript">
-  const system = <!-- system json goes here -->;
+  var system = <!-- system json goes here -->;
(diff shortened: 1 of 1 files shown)


erikfrey:
🎉🎉🎉  Thanks for submitting our very first public PR!

PR #7, opened 2021-06-25
Add __init__.py to brax subdirectories

danielsuo:
And update version number. Holding off on pushing to PyPI until we're happy.

Tested by building via `python setup.py sdist`, installing to a new environment, running our `deluca` examples and `pytest ./brax/tests -n 36`. All tests pass except the `es` ones. But those fail for both the source install and the packaged install.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ from setuptools import setup
 
 setup(
     name="brax",
-    version="0.0.2",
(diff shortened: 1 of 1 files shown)


danielsuo:
#5

erikfrey:
🙏  thank you!

PR #12, opened 2021-07-17
Add dependency management, auto version based on git tags, and auto pypi publishing

vwxyzjn:
Hello maintainers of Brax,

I really like your repo and would like to add tools to help with reproducing work in the future.

There are three parts of this PR. 

## Dependency management via `poetry`

`poetry` locks dependencies for better reproducibility. Another benefit is that I could pin the training related dependencies as dev dependencies, so that if the users (myself as an example) wants to use other DL framework to do training, they wouldn't have to also install `tensorflow` and other libraries unrelated to the env.

## Auto version based on git tags via `poetry-dynamic-versioning`
So if you run `poetry build`, it's automatically going to create distributions like `brax-0.0.0.post34.dev0+412962f.tar.gz`. And if you draft a release / create a git tag like `v0.0.1`, it's going to create a distribution like `brax-0.0.1`. So this saves a lot of time from manually incrementing `setup.py`'s versions.

## Auto PyPi publishing,
If you add a PyPi API token to the repo's secret as follows, then every time you draft a release on Github, it's automatically going to push the corresponding release to PyPi.

image

These are the best practices I learned the hard way with my own env [gym-microrts](https://github.com/vwxyzjn/gym-microrts), which also started with something like `pip install -e .`  And at some point I had the most painful time trying to reproduce some past results, which resulted in the practices I am recommending to you now :) 

For reference, I think the following snippet would work in Linux or Windows (not quite on Mac due to an upstream bug):
```bash
pip install gym_microrts==0.4.2 # make sure you have Java 8.0+ installed
curl -OL https://raw.githubusercontent.com/vwxyzjn/gym-microrts/4a8098811fc871a456041b3357f284359c466572/hello_world.py
python hello_world.py
```

I hope this might be useful to you and wish the best for your project.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -27,4 +27,4 @@ jobs:
       run: pip install poetry-dynamic-versioning && poetry install --no-dev && poetry build --format sdist
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
-      run: poetry config pypi-token.pypi ${{ secrets.pypi_token }} && poetry publish
+      run: poetry config pypi-token.pypi ${{ secrets.PYPI_API_TOKEN }} && poetry publish
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -27,4 +27,4 @@ jobs:
       run: pip install poetry-dynamic-versioning && poetry install --no-dev && poetry build --format sdist
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
-      run: poetry config pypi-token.pypi ${{ secrets.pypi_password }} && poetry publish
+      run: poetry config pypi-token.pypi ${{ secrets.pypi_token }} && poetry publish
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -1,0 +1,30 @@
+name: build
+
+on:
+  push:
+    paths-ignore:
+      - 'README.md'
+      - 'docs/**'
+      - 'notebooks/**'
+jobs:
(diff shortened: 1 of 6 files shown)


vwxyzjn:
@googlebot I signed it!

erikfrey:
Hi Costa,

Thanks so much for opening this PR.  Having the pypi push connected to a git tag looks very handy, and I agree it would be nice to have optional dependencies like tensorflow for the learners.

To fit in with other google projects our ecosystem, rather than rely on poetry we will use `bumpversion` for the versioning support, and the `extras_require` that lives in `setuptools` for optional dependencies.

It's probably disappointing to hear this as your PR is fantastic!  `poetry` just doesn't fit some of our other workflows.  If you'd like to still make sure we get this packaging functionality in yourself, we'd be very happy to work with you on a new PR and I can show you a few pointers.

Otherwise, no sweat - we'll add this functionality in the coming weeks.  Thanks again!

vwxyzjn:
Hi Erik,

Thanks for the prompt reply. Having tools that fit seamlessly with existing tool chains is definitely important! If possible, I would like to help contribute (which also helps my selfish dream to deprecate mujoco haha). If you could show me a few pointers, I could prepare another PR with `bumpversion` and others :)

erikfrey:
Hi Costa - we ended up handling automating our release flow as it was something I had to learn myself, to be honest     :-).    so I would have done a poor job at requesting or understanding changes from others.

It's not too magical - you can see there's a github action now, like you had proposed, for pushing to pypi, and some of the versioning is now handled by internal tooling.  Thanks again for this PR - I appreciate that you jumped right in!

vwxyzjn:
Hi @erikfrey, this is awesome! Thank you!

PR #13, opened 2021-07-20
Remove some unused imports

kngwyu:
Thanks for this cool project.

Here's just a nitpicking PR, which I found by running `grep -R tensorflow ./brax` (I just wanted to know what tensorflow is used for in this project).


erikfrey:
Thank you!

PR #17, opened 2021-07-22
Remove tensorflow dependency for the env

vwxyzjn:
## Description

This PR makes the use of `tensorflow.io.gfile` optional: if `tensorflow` is available, it's going to use `gfile`, otherwise it would use the default python file system to save files.

I tested `html.save_html` with both `gfile` and regular `open` and they both work.

```
import jax
import brax
import sys
from brax import envs
from brax.io import html
env_name = "ant"  # @param ['ant', 'humanoid', 'fetch', 'grasp', 'halfcheetah', 'ur5e', 'reacher']
env = envs.create_gym_env(env_name=env_name)
state = env.reset()

print(html.render(env._environment.sys, [env._state.qp]))
html.save_html("test.html", env._environment.sys, [env._state.qp])
```

### DIFF
--- a/brax/io/__init__.py
+++ b/brax/io/__init__.py
@@ -12,3 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+GFile = None
+try:
+  from tensorflow.io.gfile import GFile
+except:
+  pass
+
+class File:
+  def __init__ (self, fileName, mode='r'):
+    if GFile is not None:
+      self.f = GFile(fileName, mode)
+    else:
+      self.f = open(fileName)
+  def __enter__ (self):
+    return self.f
+  def __exit__ (self, exc_type, exc_value, traceback):
+    self.f.close()
--- a/brax/io/__init__.py
+++ b/brax/io/__init__.py
@@ -23,7 +23,7 @@ class File:
     if GFile is not None:
       self.f = GFile(fileName, mode)
     else:
-      self.f = open(fileName)
+      self.f = open(fileName, mode)
   def __enter__ (self):
     return self.f
   def __exit__ (self, exc_type, exc_value, traceback):
--- a/brax/io/file.py
+++ b/brax/io/file.py
@@ -11,3 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+
+class File:
+
+  def __init__ (self, fileName, mode='r'):
+    self.f = open(fileName, mode)
+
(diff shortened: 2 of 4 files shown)


brax/io/__init__.py
```diff
@@ -12,3 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+GFile = None
+try:
+  from tensorflow.io.gfile import GFile
+except:
+  pass
+
+class File:
```
erikfrey, line 21: please move this to a separate file.py

brax/io/__init__.py
```diff
@@ -12,3 +12,19 @@
...
+  def __init__ (self, fileName, mode='r'):
```
erikfrey, line 22: To follow the code style in this rep, can you add another newline before each method?

brax/io/__init__.py
```diff
@@ -12,3 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+GFile = None
+try:
+  from tensorflow.io.gfile import GFile
```
erikfrey, line 17: Good news!  We are removing tensorflow as a dependency here and elsewhere.  So you can omit this entirely.  I know it then raises the question as to why even have this File abstraction - we need it for some internal use cases.

So it's great if File just wraps the python file object and that's it.

erikfrey (changes requested):
Thanks for this!  And good timing for opening this PR - we are removing the other tf dependencies to make brax cleaner to import.  So thanks for adding this - just a few things to adjust.

vwxyzjn:
Hi Erik,

That's fantastic. I have just pushed changes to address your comments :)

brax/io/__init__.py
vwxyzjn: done

brax/io/__init__.py
vwxyzjn: done

brax/io/__init__.py
vwxyzjn: done

brax/io/__init__.py
```diff
@@ -11,4 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
erikfrey, line 14: You can just revert this file entirely.

erikfrey (changes requested):
Thanks so much!  One last nit and we are good to go.

brax/io/__init__.py
erikfrey: Merging as is!

PR #20, opened 2021-07-26
Added a VectorizedGymWrapper

DavidSlayback:
One of the main advantages of Brax is being able to create many parallel instances of physics environments. The original GymWrapper lets users use single-instance Brax environments in the familiar Gym API. This proposed change provides a vectorized version that follows the Gym VectorEnv API.

### DIFF
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -57,3 +57,16 @@ def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
 def create_gym_env(env_name: str, **kwargs) -> gym.Env:
   """Creates a Gym Env with a specified brax system."""
   return wrappers.GymWrapper(create(env_name, **kwargs))
+
+
+def create_gym_vector_env(env_name: str, **kwargs) -> gym.vector.VectorEnv:
+    """Creates a Gym Vector Env with a specified brax system."""
+    return wrappers.VecGymWrapper(create(env_name, **kwargs))
+
+
+def create_baselines_vec_env(env_name: str, **kwargs):
+    """Creates a StableBaselines3 VecEnv with a specified brax system."""
+    try:
+        return wrappers.VecEnvWrapper(create(env_name, **kwargs))
+    except ImportError:
+        raise ImportError('StableBaselines3 is not available.')
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -15,7 +15,7 @@
 """Some example environments to help get started quickly with brax."""
 
 import functools
-from typing import Callable
+from typing import Callable, Union, overload
 
 import gym
 import brax
@@ -54,19 +54,21 @@ def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
   return functools.partial(create, env_name, **kwargs)
 
 
-def create_gym_env(env_name: str, **kwargs) -> gym.Env:
-  """Creates a Gym Env with a specified brax system."""
-  return wrappers.GymWrapper(create(env_name, **kwargs))
+@overload
+def create_gym_env(env_name: str, seed: int = 0, **kwargs) -> wrappers.GymWrapper:
+    ...
 
 
-def create_gym_vector_env(env_name: str, **kwargs) -> gym.vector.VectorEnv:
-    """Creates a Gym Vector Env with a specified brax system."""
-    return wrappers.VecGymWrapper(create(env_name, **kwargs))
+@overload
+def create_gym_env(env_name: str, batch_size: int, seed: int = 0, **kwargs) -> wrappers.VectorGymWrapper:
+    ...
 
 
-def create_baselines_vec_env(env_name: str, **kwargs):
-    """Creates a StableBaselines3 VecEnv with a specified brax system."""
-    try:
-        return wrappers.VecEnvWrapper(create(env_name, **kwargs))
-    except ImportError:
-        raise ImportError('StableBaselines3 is not available.')
+def create_gym_env(
+    env_name: str, batch_size: int = 0, seed: int = 0, **kwargs
+) -> Union[wrappers.GymWrapper, wrappers.VectorGymWrapper]:
+    if batch_size:
+        return wrappers.VectorGymWrapper(create(env_name=env_name, batch_size=batch_size, **kwargs), seed=seed)
+    else:
+        return wrappers.GymWrapper(create(env_name=env_name, **kwargs), seed=seed)
+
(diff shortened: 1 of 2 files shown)


vwxyzjn:
Oh, I really like this change. Is this `VectorizedGymWrapper is` compatible with `openai/baselines` styled vectorized env? Are there any libraries that are built to use the `VectorizedGymWrapper` API?

DavidSlayback:
> Oh, I really like this change! Is this `VectorizedGymWrapper is` compatible with `openai/baselines` styled vectorized env? Are there any libraries that are built to use the `VectorizedGymWrapper` API?

Not directly, but the only big difference is that the Baselines VecEnv use the store just the single observation and action spaces, while the Gym.vector version keeps a batched space (easier to sample from) in addition. It wouldn't be too hard to address that and maybe make separate "create_env" functions for them

vwxyzjn:
Would you mind adding this `Baselines VecEnv` adapter? This will make it much easier to use the environments with libraries such as Stable-baselines 3. I'm also happy to build an integration to display the brax HTML visualization in the Weights and Biases dashboard (e.g. like [this](https://wandb.ai/wandb/sb3/runs/x4x9hdm2?workspace=user-costa-huang)), allowing quicker insights to see the agent's behavior in different stages of the training.

vwxyzjn:
Also on a related note, it might be tricky to use other DL learners that are not implemented in Jax. 

I was testing out the gym integration and tried to make it work with pytorch using GPU, and it had similar (slow) throughputs like the pybullet envs. Then, I realized it was probably because the flow is like `jax gpu arrays -> jax cpu arrays -> pytorch's gpu tensors`, which negates any speed advantages of brax. 

I think there's some issues tracking this to make the transfer much easier such as https://github.com/google/jax/issues/1865 and https://github.com/google/jax/issues/1100, but overall I'm not sure how to resolve this.

DavidSlayback:
I could definitely do that, though I'm not really affiliated with the team, I'm just using Brax as an alternative to Mujoco in my own research and figured it was an easy feature to quickly code up without seriously changing the codebase! 

As far as speed, this commit is specifically made to alter the behavior from the original gym as little as possible. In my own repo with pytorch, I do the jit with a cpu backend in both step and reset (to avoid the first gpu -> cpu transfer). Still better than Mujoco/Pybullet!

 Having listened to a lot of the Pytorch Dev podcast, I'm not optimistic for ever seeing a way to directly convert between PyTorch and JAX GPU tensors, there's a lot under the hood :(

DavidSlayback:
I made a very basic implementation of the Baselines wrapper you suggested. I didn't want to add the dependency, so it's wrapped in a try-except, and while many of the methods I didn't implement aren't relevant (e.g., "step_async, step_wait"), there are some ignored methods like render that could be useful.

brax/envs/wrappers.py
```diff
@@ -55,3 +56,119 @@ def reset(self):
...
+  def step(self, action):
+    self._state, obs, reward, done = self._step(self._state, action)
+    return obs, reward, done, {}
+
+  def seed(self, seed: int = 0):
+    self._key = jax.random.PRNGKey(seed)
+
+
+try:
+  from stable_baselines3.common.vec_env import VecEnv
```
lebrice, line 110: I don't feel like this belongs here, this should be done in the SB3 repo, in my opinion.

brax/envs/__init__.py
```diff
@@ -57,3 +57,16 @@ def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
 def create_gym_env(env_name: str, **kwargs) -> gym.Env:
   """Creates a Gym Env with a specified brax system."""
   return wrappers.GymWrapper(create(env_name, **kwargs))
+
+
+def create_gym_vector_env(env_name: str, **kwargs) -> gym.vector.VectorEnv:
```
erikfrey, line 62: Rather than create a new function, I think we can just check the batch_size parameter in `create_gym_env` and return `VecGymWrapper` if it's > 0

brax/envs/wrappers.py
erikfrey: +1 - also given it's only partially implemented, prefer to leave this out.  Thanks!

brax/envs/__init__.py
```diff
@@ -57,3 +57,16 @@ def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
...
+    """Creates a Gym Vector Env with a specified brax system."""
+    return wrappers.VecGymWrapper(create(env_name, **kwargs))
+
+
+def create_baselines_vec_env(env_name: str, **kwargs):
```
erikfrey, line 67: Let's avoid this for now - would rather more fully support this integration after we get more requests to support it, so that we know it's worth it.

brax/envs/wrappers.py
```diff
@@ -55,3 +56,119 @@ def reset(self):
   def step(self, action):
     self._state, obs, reward, done = self._step(self._state, action)
     return obs, reward, done, {}
+
+  def seed(self, seed: int = 0):
```
erikfrey, line 60: Thanks for adding this method - please call it from the initializer like you do in the other class.

brax/envs/wrappers.py
```diff
@@ -55,3 +56,119 @@ def reset(self):
...
+    self._key = jax.random.PRNGKey(seed)
+
+
+class VecGymWrapper(gym.vector.VectorEnv):
```
erikfrey, line 64: prefer `VectorGymWrapper` just to keep parity with naming conventions.

erikfrey (changes requested):
This is a nice improvement - thanks for submitting it!  Just left a host of nits, after which we'll be good to go.  Cheers.

brax/envs/__init__.py
lebrice: Might I suggest this:
```python3
from typing import overload

@overload
def create_env(env_name: str, **kwargs) -> GymWrapper:
    ...

@overload
def create_env(env_name: str, batch_size: int, **kwargs) -> VecGymWrapper:
    ...

def create_env(
    env_name: str, batch_size: int = None, **kwargs
) -> Union[GymWrapper, VecGymWrapper]:
    if batch_size is None:
        return create_gym_env(env_name=env_name, **kwargs)
    else:
        return create_gym_vector_env(env_name=env_name, batch_size=batch_size, **kwargs)
```

DavidSlayback:
I removed the draft work on the Baselines VecEnv and updated in response to the comments! As a side note, nice work on the torch and backend wrappers @lebrice ! I thought about including the backend as an argument here but deferred to your PR.

erikfrey (approved):
Looks great, thank you!

PR #23, opened 2021-08-02
Add wrapper for easy conversion to/from PyTorch Tensors

lebrice:
- Add JaxToTorchWrapper in to_torch.py, which converts DeviceArrays into
  PyTorch Tensors without transfering them to CPU.
- Add "backend" argument to `GymWrapper` constructor.

Signed-off-by: Fabrice Normandin <fabrice.normandin@gmail.com>

### DIFF
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -31,14 +31,14 @@ from brax.envs import ur5e
 from brax.envs import wrappers
 
 _envs = {
-  'fetch': fetch.Fetch,
-  'ant': ant.Ant,
-  'grasp': grasp.Grasp,
-  'halfcheetah': halfcheetah.Halfcheetah,
-  'humanoid': humanoid.Humanoid,
-  'ur5e': ur5e.Ur5e,
-  'reacher': reacher.Reacher,
-  'reacherangle': reacherangle.ReacherAngle,
+    'fetch': fetch.Fetch,
+    'ant': ant.Ant,
+    'grasp': grasp.Grasp,
+    'halfcheetah': halfcheetah.Halfcheetah,
+    'humanoid': humanoid.Humanoid,
+    'ur5e': ur5e.Ur5e,
+    'reacher': reacher.Reacher,
+    'reacherangle': reacherangle.ReacherAngle,
 }
 State = env.State
 Env = env.Env
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -31,32 +31,34 @@ from brax.envs import ur5e
 from brax.envs import wrappers
 
 _envs = {
-    'fetch': fetch.Fetch,
-    'ant': ant.Ant,
-    'grasp': grasp.Grasp,
-    'halfcheetah': halfcheetah.Halfcheetah,
-    'humanoid': humanoid.Humanoid,
-    'ur5e': ur5e.Ur5e,
-    'reacher': reacher.Reacher,
-    'reacherangle': reacherangle.ReacherAngle,
+    "fetch": fetch.Fetch,
+    "ant": ant.Ant,
+    "grasp": grasp.Grasp,
+    "halfcheetah": halfcheetah.Halfcheetah,
+    "humanoid": humanoid.Humanoid,
+    "ur5e": ur5e.Ur5e,
+    "reacher": reacher.Reacher,
+    "reacherangle": reacherangle.ReacherAngle,
 }
 State = env.State
 Env = env.Env
 
 
 def create(env_name: str, **kwargs) -> Env:
-  """Creates an Env with a specified brax system."""
-  return _envs[env_name](**kwargs)
+    """Creates an Env with a specified brax system."""
+    return _envs[env_name](**kwargs)
 
 
 def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
-  """Returns a function that when called, creates an Env."""
-  return functools.partial(create, env_name, **kwargs)
+    """Returns a function that when called, creates an Env."""
+    return functools.partial(create, env_name, **kwargs)
 
 
-def create_gym_env(env_name: str, seed: int = 0, **kwargs) -> gym.Env:
-  environment = create(env_name=env_name, **kwargs)
-  if environment.batch_size:
-    return wrappers.VectorGymWrapper(environment, seed=seed)
-  else:
-    return wrappers.GymWrapper(environment, seed=seed)
+def create_gym_env(
+    env_name: str, seed: int = 0, backend: str = "cpu", **kwargs
+) -> gym.Env:
+    environment = create(env_name=env_name, **kwargs)
+    if environment.batch_size:
+        return wrappers.VectorGymWrapper(environment, seed=seed, backend=backend)
+    else:
+        return wrappers.GymWrapper(environment, seed=seed, backend=backend)
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -31,80 +31,80 @@ from brax.envs import ur5e
 from brax.envs import wrappers
 
 _envs = {
-    "fetch": fetch.Fetch,
-    "ant": ant.Ant,
-    "grasp": grasp.Grasp,
-    "halfcheetah": halfcheetah.Halfcheetah,
-    "humanoid": humanoid.Humanoid,
-    "ur5e": ur5e.Ur5e,
-    "reacher": reacher.Reacher,
-    "reacherangle": reacherangle.ReacherAngle,
+  'fetch': fetch.Fetch,
+  'ant': ant.Ant,
+  'grasp': grasp.Grasp,
+  'halfcheetah': halfcheetah.Halfcheetah,
+  'humanoid': humanoid.Humanoid,
+  'ur5e': ur5e.Ur5e,
+  'reacher': reacher.Reacher,
+  'reacherangle': reacherangle.ReacherAngle,
 }
 State = env.State
 Env = env.Env
 
 
 def create(env_name: str, **kwargs) -> Env:
-    """Creates an Env with a specified brax system."""
-    return _envs[env_name](**kwargs)  # type: ignore
+  """Creates an Env with a specified brax system."""
+  return _envs[env_name](**kwargs)  # type: ignore
 
 
 def create_fn(env_name: str, **kwargs) -> Callable[..., Env]:
-    """Returns a function that when called, creates an Env."""
-    return functools.partial(create, env_name, **kwargs)
+  """Returns a function that when called, creates an Env."""
+  return functools.partial(create, env_name, **kwargs)
 
 
 @overload
 def create_gym_env(
-    env_name: str,
-    batch_size: None = None,
-    seed: int = 0,
-    backend: Optional[str] = None,
-    **kwargs
+  env_name: str,
+  batch_size: None = None,
+  seed: int = 0,
+  backend: Optional[str] = None,
+  **kwargs
 ) -> gym.Env:
-    ...
+  ...
 
 
 @overload
 def create_gym_env(
-    env_name: str,
-    batch_size: int,
-    seed: int = 0,
-    backend: Optional[str] = None,
-    **kwargs
+  env_name: str,
+  batch_size: int,
+  seed: int = 0,
+  backend: Optional[str] = None,
+  **kwargs
 ) -> gym.vector.VectorEnv:
-    ...
+  ...
 
 
 def create_gym_env(
-    env_name: str,
-    batch_size: Optional[int] = None,
-    seed: int = 0,
-    backend: Optional[str] = None,
-    **kwargs
+  env_name: str,
+  batch_size: Optional[int] = None,
+  seed: int = 0,
+  backend: Optional[str] = None,
+  **kwargs
 ) -> Union[gym.Env, gym.vector.VectorEnv]:
-    """Creates a `gym.Env` or `gym.vector.VectorEnv` from a Brax environment.
+  """Creates a `gym.Env` or `gym.vector.VectorEnv` from a Brax environment.
 
-    Parameters
-    ----------
-    env_name : str
-        Name of the environment to create.
-    batch_size : Optional[int], optional
-        Number of parallel environments. Defaults to `None`, in which case a single env
-        is returned. When `batch_size` > 1, a subclass of `gym.vector.VectorEnv` is
-        returned instead.
-    seed : int, optional
-        Random seed, by default 0.
-    backend : str, optional
-        Backend used for jit compilation of the `reset` and `step` methods. Defaults to
-        `None`, in which case the backend is chosen automatically.
+  Parameters
+  ----------
+  env_name : str
+    Name of the environment to create.
+  batch_size : Optional[int], optional
+    Number of parallel environments. Defaults to `None`, in which case a single env
+    is returned. When `batch_size` > 1, a subclass of `gym.vector.VectorEnv` is
+    returned instead.
+  seed : int, optional
+    Random seed, by default 0.
+  backend : str, optional
+    Backend used for jit compilation of the `reset` and `step` methods. Defaults to
+    `None`, in which case the backend is chosen automatically.
 
-    Returns
-    -------
-    Union[gym.Env, gym.vector.VectorEnv]
-        A `gym.Env` or a gym.vector.VectorEnv`, depending on the value of `batch_size`.
-    """
-    environment = create(env_name=env_name, batch_size=batch_size, **kwargs)
-    if batch_size:
-        return wrappers.VectorGymWrapper(environment, seed=seed, backend=backend)
-    return wrappers.GymWrapper(environment, seed=seed, backend=backend)
+  Returns
+  -------
+  Union[gym.Env, gym.vector.VectorEnv]
+      A `gym.Env` or a gym.vector.VectorEnv`, depending on the value of `batch_size`.
+  """
+  environment = create(env_name=env_name, batch_size=batch_size, **kwargs)
+  if batch_size:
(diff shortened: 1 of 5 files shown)


lebrice:
@googlebot  I signed it!

erikfrey:
Thanks for proposing this - I think it's going to be helpful to a lot of other users.  I'll do a review of this PR in the next few days.

brax/envs/__init__.py
```diff
@@ -30,33 +30,79 @@
...
-_envs = {
-    'fetch': fetch.Fetch,
-    'ant': ant.Ant,
-    'grasp': grasp.Grasp,
-    'halfcheetah': halfcheetah.Halfcheetah,
-    'humanoid': humanoid.Humanoid,
-    'ur5e': ur5e.Ur5e,
-    'reacher': reacher.Reacher,
-    'reacherangle': reacherangle.ReacherAngle,
+_envs: Dict[str, Type[env.Env]] = {
```
erikfrey, line 33: To be consistent with the rest of the library, please avoid type annotations for internal constants.

brax/envs/__init__.py
```diff
@@ -30,33 +30,79 @@
...
+    "reacherangle": reacherangle.ReacherAngle,
 }
 State = env.State
 Env = env.Env
 
 
 def create(env_name: str, **kwargs) -> Env:
-  """Creates an Env with a specified brax system."""
-  return _envs[env_name](**kwargs)
+    """Creates an Env with a specified brax system."""
```
erikfrey, line 48: Nit: It looks like you're using some kind of linter with a different code style than ours - changes like this will be reverted by our internal tooling.  So for a cleaner git history, please avoid style changes re: tab widths, replacing ' with ", and so on.

brax/envs/__init__.py
```diff
@@ -30,33 +30,79 @@
...
+    return functools.partial(create, env_name, **kwargs)
 
 
-def create_gym_env(env_name: str, seed: int = 0, **kwargs) -> gym.Env:
-  environment = create(env_name=env_name, **kwargs)
-  if environment.batch_size:
-    return wrappers.VectorGymWrapper(environment, seed=seed)
-  else:
-    return wrappers.GymWrapper(environment, seed=seed)
+@overload
```
erikfrey, line 57: Our internal type checking system complained about the `@overload` annotations... I'm not entirely sure why, but I suspect it's something to do with kwargs making the different function signatures ambiguous for matching.

In any case, I think we gain some simplicity by explicitly only returning `gym.Env` as the API, rather than exposing the implementation types `GymWrapper` and `VectorGymWrapper` to the user.

Can you see if your PR works without these overloads, and keeping the old single `create_gym_env(...)-> gym.Env` function signature?  But go ahead and add the `backend` parameter - that is useful, thanks.

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
+""" Wrapper around a Brax GymWrapper, that converts its outputs to PyTorch tensors.
+
+This conversion happens direcly on-device, witout the need to move values to the CPU.
```
erikfrey, line 3: nit: directly

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+        action = self.action(action)
+        obs, rewards, done, info = super().step(action)
+        obs = self.observation(obs)
+        rewards = self.reward(rewards)
+        done = self.done(done)
+        info = self.info(info)
+        return obs, rewards, done, info
+
+
+@singledispatch
```
erikfrey, line 76: These functions are cool and may be generally useful elsewhere!  Can you move them into something like `brax/io/torch.py`, and then put the wrapper into `wrappers.py`?

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+            A `GymWrapper` or `VectorGymWrapper` to wrap.
+        device : torch.device, optional
+            device on which to move the Tensors. Defaults to `None`, in which case the
+            tensors will be on the same devices as their Jax equivalents.
+        """
+        if not hasattr(env, "reward_range"):
+            # NOTE: Fix a bug with the current GymWrapper, where the `Env` doesn't have
+            # the `reward_range` attribute which is needed by the gym.Wrapper
+            # constructor.
+            env.reward_range = (-np.inf, np.inf)
```
erikfrey, line 43: If `GymWrapper` and `VectorGymWrapper` are missing an important attribute, by all means, add them there!  Thanks.

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+import datetime
+import time
+from collections import abc
+from functools import partial, singledispatch
+from typing import Any, Dict, Mapping, Optional, TypeVar, Union
+
+import gym
+import matplotlib.pyplot as plt
+import numpy as np
+import torch
```
erikfrey, line 15: Much like we are striving to make tensorflow an optional dependency, let's do the same here.  To that end, can you try something like this:

https://github.com/google/brax/blob/main/brax/io/export.py#L24

Just so the user gets a more informative message if the import fails.  Thanks.

brax/envs/__init__.py
```diff
@@ -30,33 +30,79 @@
...
+def create_gym_env(
+    env_name: str,
+    batch_size: None = None,
+    seed: int = 0,
+    backend: str = "cpu",
```
erikfrey, line 62: Given that these changes are making the gym Env much more flexible/powerful, I think it's okay to default this to `None` now to allow the default behavior of automatic device selection.

brax/envs/wrappers.py
```diff
@@ -20,12 +20,17 @@
...
 
 class GymWrapper(gym.Env):
   """A wrapper that converts Brax Env to one that follows Gym API."""
 
-  def __init__(self, environment: env.Env, seed: int = 0):
+  # Flag that prevents `gym.register` from misinterpreting the `_step` and `_reset` as
+  # signs of a deprecated gym Env API.
+  _gym_disable_underscore_compat: ClassVar[bool] = True
+
+  def __init__(self, environment: env.Env, seed: int = 0, backend: str = "cpu"):
```
erikfrey, line 33: same here: can default to None

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+"""
+
+import datetime
+import time
+from collections import abc
+from functools import partial, singledispatch
+from typing import Any, Dict, Mapping, Optional, TypeVar, Union
+
+import gym
+import matplotlib.pyplot as plt
```
erikfrey, line 13: nit: unneeded import?

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+    def info(self, info: Dict) -> Dict:
+        return jax_to_torch(info, device=self.device)
+
+    def reset(self):
+        obs = super().reset()
+        return self.observation(obs)
+
+    def step(self, action):
+        action = self.action(action)
+        obs, rewards, done, info = super().step(action)
```
erikfrey, line 68: nit: reward

brax/envs/to_torch.py
```diff
@@ -0,0 +1,133 @@
...
+        obs = self.observation(obs)
+        rewards = self.reward(rewards)
```
erikfrey, line 70: nit: reward

erikfrey (changes requested):
This is really great!  Do you have an example colab that ties this all together?  If so, I'd be happy to incorporate it into `notebooks` and link to it in the README.  I could see this being useful for a lot of folks.  Nice work!

brax/envs/__init__.py
lebrice: I see. I use black. Which auto-formatting tool do you recommend?

brax/envs/__init__.py
lebrice: Would a `Union[gym.Env, gym.vector.VectorEnv]` work?
What kind of error are you getting? Might I ask what type checking system you are using? Is it comparable to mypy? 
It is in my opinion somewhat important for this function to indicate the distinction, since they behave differently, in subtle yet important ways! (e.g. `done` being a vector, the presence of `single_observation_space`, `num_envs`, etc.)

brax/envs/to_torch.py
lebrice: I agree! I will set this as the default value in those classes for now. 
However, I wouldn't know how to determine that value from the wrapped Brax env, do you have any idea on how I could get the min reward and max rewards from it?

brax/envs/to_torch.py
lebrice: > an you move them into something like brax/io/torch.py

Absolutely!

>  and then put the wrapper into wrappers.py?

I don't think this is a good idea: it would make `torch` a mandatory requirement for using anything in `wrappers.py`!

brax/envs/to_torch.py
lebrice: Nevermind, it's actually not needed, since the `gym.Env` class has a default value for it. I'm not sure why I was getting an error before :monocle_face:

lebrice:
> This is really great! Do you have an example colab that ties this all together? If so, I'd be happy to incorporate it into notebooks and link to it in the README. I could see this being useful for a lot of folks. Nice work!

Sure! I've added it under `notebooks/gym_pytorch.ipynb`. Would you like me to keep it as part of this PR, or create another one for it?

brax/envs/to_torch.py
lebrice: Perhaps there could be a `wrappers` folder that would contain the contents of `wrappers.py`, renamed to something like `base.py`, and then this `to_torch.py` file would be placed in there? What do you think?
This would be done in another PR if you want.

brax/envs/__init__.py
erikfrey: I actually recommend avoiding auto-formatting!  This is formatted by internal Google tools that isn't public.  I'm sure you could probably customize an auto-formatter to better match this format, if that's helpful to you.

But the general principle I'd ask you to adhere to for now is to help us keep the blame history clean by avoiding changes that are unrelated to the PR, such as whole-file style changes that come from using an auto-formatter.

In the future as the project grows we could find some acceptable code style from something like yapf: https://github.com/google/yapf - I'll add an issue for it.

brax/envs/__init__.py
erikfrey: Avoiding the `@overload` and returning `Union[gym.Env, gym.vector.VectorEnv]` would fix the specific error I saw, yes. 

Where do you use `single_observation_space` or `num_envs`?  I don't see them accessed anywhere.  This is actually a good thing, as it means you are conforming to the gym API exactly, and that you don't need extra hidden information to process the environment.  If you do need to access them, you can just 

Looking through this code again, I would actually recommend removing them as fields.  And if you do need `num_envs` or `single_observation_space`, you can get both directly from `observation_space`, no?

The concern that `done` changes from a scalar to a vector I'm sympathetic to, but I think the `gym.Env` api already has a rich history of returning very different types depending on how it is instantiated, with no type guarantees.  Unless there's a use case right now that breaks, I think it's OK to follow this same convention.

brax/io/torch.py
```diff
@@ -0,0 +1,82 @@
...
+from collections import abc
+
+from jax._src import dlpack as jax_dlpack
+from jaxlib.xla_extension import DeviceArray
+
+try:
+    import torch
+except ImportError:
+    warnings.warn(
+        "to_torch requires PyTorch. Please run `pip install torch` for this function "
```
erikfrey, line 15: nit: what is to_torch?  Did you mean to say maybe:

```
"brax.io.torch requires PyTorch.  Please run `pip install torch` for this module to work."
```

notebooks/gym_pytorch.ipynb
```diff
@@ -0,0 +1,222 @@
...
+    },
+    {
+     "output_type": "stream",
+     "name": "stdout",
+     "text": [
+      "Device used: cuda:0\n",
+      "Number of parallel environments: 1024\n",
+      "Average time per batched step:  0.0034599435329437257 ± 0.00036434224775974217 seconds\n",
+      "Frequency (after first two steps): ~289.022 batched steps / second.\n",
+      "Effective Frequency (after first two steps): ~295958.587 steps / second.\n"
```
erikfrey, line 184: 🎉

notebooks/gym_pytorch.ipynb
```diff
@@ -0,0 +1,222 @@
+{
```
erikfrey, line 1: Awesome, thank you!  We may mangle this a bit to fit it into a larger notebook reorg that's coming in a bit, but this is already a really helpful example.

erikfrey (changes requested):
Getting close... thanks Fabrice.

brax/envs/to_torch.py
erikfrey: Oops - I missed this.  Yes, good point.  We can save that for some other time - leaving it as is for now works for me.

brax/envs/__init__.py
lebrice: > Avoiding the @overload and returning Union[gym.Env, gym.vector.VectorEnv] would fix the specific error I saw, yes.

What was that error? Would you be able to share the error message you were getting?

> Where do you use single_observation_space or num_envs? I don't see them accessed anywhere. This is actually a good thing, as it means you are conforming to the gym API exactly, and that you don't need extra hidden information to process the environment.

What i mean is that it's important for the user code to be able to tell (at static type-checking time) whether the returned value is a `gym.Env` instance, or a `gym.vector.VectorEnv` instance, because if it's a `gym.vector.VectorEnv`, then it behaves differently than a `gym.Env`.
Sorry I wasn't clear before.

notebooks/gym_pytorch.ipynb
lebrice: By the way, I'll change the branch name in the colab sticker to point to main if that's alright with you.

brax/envs/__init__.py
lebrice: > I actually recommend avoiding auto-formatting! 

I couldn't disagree more! :stuck_out_tongue: But that's a conversation for another time.

Good idea for the issue. I recommend `black`, or at least something completely unambiguous.

lebrice: Alright I've reverted the formatting changes to `brax/envs/__init__.py`.

brax/envs/__init__.py
erikfrey: There error was from using pytype:

```
File "brax/envs/__init__.py", line 78, in create_gym_env: bad return type [bad-return-type]
           Expected: gym.vector.vector_env.VectorEnv
  Actually returned: brax.envs.wrappers.GymWrapper
```

See: https://google.github.io/pytype/errors.html#bad-return-type

Why is it important for the user code to distinguish between these two types statically?  I hear you that the implementations are different, but what matters is that the API of the two classes is exactly the same.  This is exactly why `gym.vector.VectorEnv` inherits from `gym.Env`, so that users can just accept `gym.Env` and move on.  What specific use case are you seeing that is breaking otherwise?

notebooks/gym_pytorch.ipynb
erikfrey: Makes sense, thanks.

erikfrey (changes requested):

brax/envs/__init__.py
erikfrey: Thanks.  Definitely want to standardize on an easy to use formatter.  This is actually being actively discussed in the jax team, and we'll probably just follow their lead.

erikfrey: So we don't forget: https://github.com/google/brax/issues/24

brax/envs/__init__.py
lebrice: Thanks for giving the error, I'll take a look and try to find a fix.

Yes, VectorEnv inherits from gym.Env, but that doesnt mean that the APIs are exactly the same, it just means that there is overlap!

Yes there are plenty of differences which are very important to be aware of from the user side! Here is but a few:

- Done is a vector
- (currently) no rendering support
- final observation is wasted in VectorEnvs
- VectorEnvs can't (currently) be vectorized

Since you can't (and shouldn't) treat aVectorEnv *exactly* like a gym.Env, it would be very good, in my opinion, that if you use a single function for both, that the static type checking can at least inform users of which one they are going to get.

If you really can't tolerate the typing error, then Id suggest reverting back to two different functions, one for a gym.Env, and another for a VectorEnv.

Returning Union, while it appears to be a bit of a middle-ground, actually makes user code uglier and more complicated, because it shifts the responsibility of checking the return type to the user side, using either asserts or isinstance checks in user code, which isnt great.

lebrice: By the way I think the error might be caused by the value batch_size=0, which is an int, so the type check expects a VectorEnv, but the if clause returns a Env instead!

I'll try a few things tomorrow, but Im pretty confident that I'll be able to get that error to disapear.

lebrice:
Hey @erikfrey, would you mind check if the type error is fixed now?
I also reverted the changes to the formatting.

erikfrey:
The error is fixed.  Thanks for this awesome improvement!

brax/io/torch.py
```diff
@@ -0,0 +1,82 @@
+""" Generic functions to convert Jax DeviceArrays into PyTorch Tensors and vice-versa.
+"""
+import warnings
+from collections import abc
+from functools import singledispatch
+from typing import Any, Union, Dict
+
+from jax._src import dlpack as jax_dlpack
+from jaxlib.xla_extension import DeviceArray
```
erikfrey, line 9: @lebrice I had to fix this internally to:

```
from jax.interpreters.xla import DeviceArray
```

Would you mind confirming that this doesn't break anything for you?

PR #30, opened 2021-08-13
Support of height maps and collision between box corner and height map

o-Oscar:
Hello maintainers of Brax,

I really like your repo and would like to add support for height maps in order to train locomotion policies in uneven terrain.

This PR implements both :
- The visualization of height maps with the THREE interface.
- The collision handling between height maps and box corners.

I hope this might be useful to you and wish the best for your project.


o-Oscar:
@googlebot I signed it!

o-Oscar:
@googlebot I fixed it.

brax/io/html.py
```diff
@@ -166,6 +166,52 @@ def render(sys: brax.System, qps: List[brax.QP]) -> str:
   return mesh;
 }
 
+function createHeightMap(heightMap) {
```
erikfrey, line 169: Maintaining these js files inside python was difficult, so we moved them out to a `js/` directory - apologies for that change conflicting with yours.  Can you move these changes to the new location?

brax/io/html.py
```diff
@@ -166,6 +166,52 @@ def render(sys: brax.System, qps: List[brax.QP]) -> str:
...
+    const idx = Math.round(v*(n_subdiv) + u*n_subdiv*(n_subdiv+1));
+    //console.debug(idx)
+    const x = u*size;
+    const y = -v*size;
+    const z = heightMap.data[idx];
+    target.set(x, y, z).multiplyScalar(1);
+  }
+
+
+  const geometry = new THREE.ParametricGeometry(builder, n_subdiv, n_subdiv);
```
erikfrey, line 192: Cool class, I didn't know about this.

brax/io/html.py
```diff
@@ -166,6 +166,52 @@ def render(sys: brax.System, qps: List[brax.QP]) -> str:
...
+      new THREE.MeshStandardMaterial( { color: 0x796049, flatShading: true} )
+    );
+  mesh.rotation.x = -Math.PI / 2;
+  mesh.receiveShadow = true;
+  group.add(mesh);
+
+  const mesh2 = new THREE.GridHelper(2000, 2000, 0x000000, 0x000000);
+  mesh2.material.opacity = 0.4;
+  mesh2.material.transparent = true;
+  //group.add(mesh2);
```
erikfrey, line 209: if you aren't using this mesh, you can just delete the whole declaration.

brax/io/html.py
```diff
@@ -176,7 +222,7 @@ def render(sys: brax.System, qps: List[brax.QP]) -> str:
   scene.add(hemiLight);
 
   const dirLight = new THREE.DirectionalLight(0xffffff);
-  dirLight.position.set(3, 10, 10);
+  dirLight.position.set(3, 5, 10);
```
erikfrey, line 225: I see a few changes: this line, the autoRotate change below, and the camera distance change below - do you have a specific scene that needs these changes?  I would suggest rather than changing them for all scenes, instead either:

1) expose them as properties in the gui
2) find clever ways of automatically determining good defaults given the scene information
3) leaving the defaults as is for now if 1) or 2) is too tricky.  We can always add an issue for determining better defaults and solve it in a different PR.

brax/physics/colliders.py
```diff
@@ -274,6 +274,119 @@ def apply(cap_a, cap_b, qp_a, qp_b):
...
     return P(vel_a + vel_b, ang_a + ang_b)
 
+class BoxHeightMap:
+  """ A collision between the box corners and a Heihtmap. """
+
+  def __init__(self, config: config_pb2.Config):
+    self.config = config
+    self.pairs = _find_body_pairs(config, 'box', 'heightMap')
+    if not self.pairs:
+      print("no_pairs")
```
erikfrey, line 284: please remove print statement

brax/physics/colliders.py
```diff
@@ -494,7 +607,8 @@ def _collide(config: config_pb2.Config, body: bodies.Body, qp: QP,
 
   # factor of 2.0 here empirically helps object grip
   # TODO: expose friction physics parameters in config
-  return dp_n * colliding_n + dp_d * colliding_d * 2.0
+  # This factor of two makes not much of a physical sense. It kinda breaks the math, and it breaks collisions with normal not alligned with gravity vector.
```
erikfrey, line 610: We have a fix for this in progress - thanks for creating the issue.  If you would like to wait for it, you can rebase after the fix is in.  It also requires changing a few hyperparams in the codebase elsewhere.

brax/physics/config.proto
```diff
@@ -61,6 +47,13 @@ message Collider {
   message Sphere {
     float radius = 1;
   }
+  // A height map aligned with the x-y plane
+  message HeightMap {
+    float size = 1;
+    int32 meshSize = 2;
```
erikfrey, line 53: it seems your code assumes the mesh is square, which is fine by me.  If that's the case, can we remove this field, and then raise an exception if len(data) is not square?

brax/physics/config.proto
```diff
@@ -113,10 +107,10 @@ message Joint {
   repeated Range angle_limit = 9;
   // Strength of the force that restores joint to within the angle limits
   // If not set, defaults to a joint-dependent fraction of `stiffness`.
-  optional float limit_strength = 10;
```
erikfrey, line 116: we are using the latest version of proto3 where optional means that, upon deserialization, you can actually check for the presence of scalars.  So this is important for us to keep.

erikfrey (changes requested):
This is probably the first example of a mesh in Brax, which is awesome!  Thanks for creating this PR and we're glad you're enjoying Brax.

Beyond the PR comments I have two asks for you:

- How hard would it be to add a simple unit test into `physics_test.py`?  The code is just tricky enough that it would be nice to have something in place so that future developers can be sure they don't break its functionality.
- How hard would it be to add capsule<>height map?  Capsules are more popular than boxes for most users (but having the box support is great!)

brax/io/html.py
o-Oscar: Those changes were only done to have a better shading on the terrain and validate the underlying geometry. I will revert those changes to default.

brax/physics/config.proto
o-Oscar: Yes, checking for squareness of len(data) and raising an exception is indeed the better choice.

o-Oscar:
Thanks a lot for the reply and the comments.

> How hard would it be to add a simple unit test into physics_test.py? The code is just tricky enough that it would be nice to have something in place so that future developers can be sure they don't break its functionality.

I was thinking about a simple test that would check that a box does not fall through a height map. Would that be good for you ?

> How hard would it be to add capsule<>height map? Capsules are more popular than boxes for most users (but having the box support is great!)

The only collision check that I implemented is boxes corner with the height map. Because a corner is just a point and can intersect with at most the only triangle that is right below it, the cost for computing those collisions is O(1).

With capsules, we would need to compute collisions with every triangle that is below the two sphere and the center cylinder. An upper bound for the number of such triangle is given by the size of the capsule divided by the size of the height map triangle. If the resolution of the height map is much smaller than the size of the capsule, the computation cost and compile time would grow quite high.

On solution would be to only handle the case of small capsule in coarse terrains.

brax/physics/colliders.py
o-Oscar: Ok, I will leave it as is. I will fill up the issue with more precise example of the behavior that arise with this factor.

brax/io/html.py
o-Oscar: Yes, moving to the new location is quite easy. 
But it now seems that testing and implementing new features on local branches might become very complicated due to how the js script is loaded : io/html.py makes a static reference to `https://cdn.jsdelivr.net/gh/google/brax@a93dadc48a41177a4d9e7794ce17c1ad98ed8583/js/viewer.js` which I believe is not easily editable by us.

erikfrey: Yes agreed, local dev is not obvious how to do.  That static reference is not available to you, but if you fork the code, you can get your own cdn url using cdnjs, and you can use `@latest` in the url so that the url remains static.  That's still a bit annoying though because you have to push each time you want to try a new version of your js.

So here's another option for quick local dev:

In a cell in the colab, start up a local webserver that is pointed at your local checkout of the brax repo:

```
import threading
import socket
import http.server
import socketserver

class V6Server(socketserver.TCPServer):
  address_family = socket.AF_INET6

class Handler(http.server.SimpleHTTPRequestHandler):
  def __init__(self, *args, **kwargs):
    super().__init__(*args, directory='/path/to/brax', **kwargs)

port = 32888

def server_entry():
    httpd = V6Server(('::', port), Handler)
    httpd.serve_forever()

thread = threading.Thread(target=server_entry)
thread.start()
```

Then in `html.io`, change the javascript url like so:

image

If this hack works for you, maybe we can introduce some kind of flag into the code to support this local dev more easily.

erikfrey:
Comments in-line:

> ... snip ... 
> I was thinking about a simple test that would check that a box does not fall through a height map. Would that be good for you ?

That test sounds perfect.

> ... snip ...
> The only collision check that I implemented is boxes corner with the height map. Because a corner is just a point and can intersect with at most the only triangle that is right below it, the cost for computing those collisions is O(1).
> 
> With capsules, we would need to compute collisions with every triangle that is below the two sphere and the center cylinder. An upper bound for the number of such triangle is given by the size of the capsule divided by the size of the height map triangle. If the resolution of the height map is much smaller than the size of the capsule, the computation cost and compile time would grow quite high.
> 
> On solution would be to only handle the case of small capsule in coarse terrains.

Interesting, OK.  I think this is exactly the same as the problems with the box<>heightmap approach, yes?  We might have a collision between a box plane and the mesh without any of the corners colliding.  I guess this is OK as long as we make this very clear in the collider code via a comment.

brax/physics/colliders.py
```diff
@@ -274,6 +274,119 @@ def apply(cap_a, cap_b, qp_a, qp_b):
...
+      dP: Delta velocity to apply to the box bodies in the collision.
+      colliding: Mask for each body: 1 = colliding, 0 = not colliding.
+    """
+    if not self.pairs:
+      return P(jnp.zeros_like(qp.vel), jnp.zeros_like(qp.ang))
+
+    @jax.vmap
+    def apply(box, corner, qp_box, qp_heightMap, size, meshSize, heights):
+      pos, vel = math.to_world(qp_box, corner)
+      uv_pos = (pos[:2]-qp_heightMap.pos[:2])/size*(meshSize-1)
```
erikfrey, line 347: Will this handle rotated height maps?  Users may want to use heightmaps as walls or ramps or some such. 
 Presumably, looking up the uv index of a rotated height map shouldn't be a problem, right?

brax/io/html.py
o-Oscar: No, it did not work for me. Although I can access it easily from the browser, the javascript import fails silently when I try to import the local file viewer.js. I will revert to the old method for dev, But this is really frustrating for local dev and for future contributors.

brax/physics/colliders.py
o-Oscar: Indeed this does not handle rotated height maps. I am changing that.

o-Oscar:
Everything done !
Height map support should be complete. It includes
- Updated config.proto for height map support
- Collision with box corners. (Collision of height-map vertices with box not supported though).
- Render in the html Brax viewer.
- Unit test that checks that a box does not fall through a flat height map.

brax/io/html.py
erikfrey: Ahh, that's a bummer that you found this new js setup frustrating to dev.  Let's fix this!  Here's a colab that demonstrates serving local js:

https://colab.sandbox.google.com/gist/erikfrey/eb8b138dce06f7c2135e9d9690a09a8e/local_dev.ipynb#scrollTo=NaJDZqhCLovU

It might be a bit different if you were actually hosting the jupyter notebook on your local machine.  I just created an issue to track this here: https://github.com/google/brax/issues/33

brax/physics/config_pb2.py
```diff
@@ -1,20 +1,6 @@
...
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
-# source: brax/physics/config.proto
```
erikfrey, line 17: Thanks for regenerating these!  I realize we don't have instructions for contributors on how to do this - I'll add them in the future.  The only issue here is that this path information is important, as it propagates down into the message descriptor properties.  So to regenerate, instead of doing it within `brax/physics/`, instead do:

```cd <my-brax-project-dir>
python3 -m grpc_tools.protoc -I . --python_out=.  brax/physics/config.proto
```

brax/physics/colliders.py
```diff
@@ -275,6 +275,125 @@ def apply(cap_a, cap_b, qp_a, qp_b):
...
+
+    @jax.vmap
+    def apply(box, corner, qp_box, qp_heightMap, size, meshSize, heights):
+      world_pos, vel = math.to_world(qp_box, corner)
+
+      pos = math.inv_rotate(world_pos-qp_heightMap.pos, qp_heightMap.rot)
+
+      uv_pos = (pos[:2])/size*(meshSize-1)
+
+      uv_idx = jnp.floor(uv_pos).astype(jnp.int32) # poorly choosen name for the indexes of the point that form the square around the corner
```
erikfrey, line 355: nit: is there a better name?     :-)

barring that, just a comment describing this algorithm might allow us to not feel bad about the quality of the name, e.g:

```
# first we find the square in the mesh that enclose the candidate point, with mesh indices ux_idx, ux_udx_u, ...
# then we subdivide the square to find the triangle, with points bla bla
```

brax/physics/colliders.py
```diff
@@ -275,6 +275,125 @@ def apply(cap_a, cap_b, qp_a, qp_b):
 
     return P(vel_a + vel_b, ang_a + ang_b)
 
+class BoxHeightMap:
+  """ A collision between the box corners and a Heihtmap. """
```
erikfrey, line 279: nit: s/Heihtmap/Heightmap

js/system.js
```diff
@@ -63,6 +63,40 @@ function createSphere(sphere, name) {
   return mesh;
 }
 
+function createHeightMap(heightMap) {
+
+  const size = heightMap.size;
+  const n_subdiv = Math.sqrt(heightMap.data.length)-1;
+
+  if (!Number.isInteger(n_subdiv)) {
+    throw 'The data lenght for an height map should be a perfect square.';
```
erikfrey, line 72: nit: s/lenght/length

brax/physics/colliders.py
```diff
@@ -275,6 +275,125 @@ def apply(cap_a, cap_b, qp_a, qp_b):
...
+            dtype=jnp.float32)
+        col = box.colliders[0]
+        corner = corner * vec_to_np(col.box.halfsize)
+        corner = math.rotate(corner, euler_to_quat(col.rotation))
+        corner = corner + vec_to_np(col.position)
+        corners.append(corner)
+
+        meshSize = int(jnp.round(jnp.sqrt(len(heightMap.colliders[0].heightMap.data))))
+        if not len(heightMap.colliders[0].heightMap.data) == meshSize ** 2:
+          raise ValueError("The data lenght for an height map should be a perfect square.")
```
erikfrey, line 312: nit: s/lenght/length

erikfrey (changes requested):
Nice!  We are down to one change and a few comments nits.  Thanks!

o-Oscar:
Thanks for being so patient !
Everything should be corrected now.

erikfrey:
🎉 Merged!  Thank you!

PR #51, opened 2021-09-02
Add to readme how to install brax from pypi

benelot:
Hello! I remember somebody telling me that brax can be installed from pypi. However, this is not documented here in the readme. Is this not the recommended way to install it? I would suggest it is, no?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,15 @@ pip install --upgrade pip
 pip install -e .
 ```
 
+Alternatively, to install Brax from pypi, install it with:
+
+```
+python3 -m venv env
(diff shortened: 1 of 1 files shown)


erikfrey:
Yes, it is probably time we let people know pypi is available     :-)

In fact I think we can switch the order and make the installing from source be the "Alternatively, if you want to..."

WDYT?

benelot:
Absolutely, this is what I meant by "is this not the recommended way". Pypi first, source for the tinkerers and developers.

erikfrey:
Looks great, thanks!

PR #59, opened 2021-09-13
Enable per-collider friction specification

### LINKED ISSUE #55: Friction per body?
Hi,

I'm trying to create multiple bodies in the environment with differing friction coefficients. However, I couldn't find anywhere how to do this. It seems that most examples have a global friction coefficient of 0.6 but no difference across bodies.

I'd appreciate any thoughts on this. Thanks.

namheegordonkim:
* Add optional float field "friction" to colliders; If unspecified, the global friction coefficient is applied
* Re-compile protobuf file (config_pb2.py)
* Introduce Collider dataclass to pass into _collide() and _collide_pair()
* Replace references to config.friction with per-body friction (Collider.friction)

Fixes #55 

Please note:

1. I did some primitive tests within `physics_test.py` but observed that the simulation is currently unstable with friction due to lateral friction, no matter how large I set the number of substeps. Hence I ended up rolling back, in case someone else can advise. I'd appreciate any help writing tests for these changes in near future.

2. I am preserving the plane-body (and mesh-body?) collision behaviour in  `_collide()`, i.e. only the body's coefficient matters and not the plane's. However, a plane's collision coefficient is pretty important for rolling and sliding motions. I'm not a simulation expert yet, but I'd love to see how e.g. Bullet or MuJoCo implements lateral friction.

### DIFF
--- a/brax/physics/colliders.py
+++ b/brax/physics/colliders.py
@@ -30,6 +30,28 @@ from brax.physics.base import P, QP, euler_to_quat, vec_to_np, take
 from brax.physics.math import safe_norm
 
 
+@struct.dataclass
+class Collider:
+  """
+  A Collider contains information pertaining to collision such as friction and elasticity.
+  """
+  idx: jnp.ndarray
+  friction: jnp.ndarray
+
+  @classmethod
+  def from_config(cls, config: config_pb2.Config) -> 'Collider':
+    """Returns Collider from a brax config."""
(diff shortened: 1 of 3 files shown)


namheegordonkim:
FYI, just signed CLA. Conflicts should be straightforward to resolve

namheegordonkim:
Will create a new PR after resolving conflicts

PR #60, opened 2021-09-13
Enable per-collider friction specification

### LINKED ISSUE #55: Friction per body?

namheegordonkim:
* Add optional float field "friction" to colliders; If unspecified, the global friction coefficient is applied
* Re-compile protobuf file (config_pb2.py)
* Introduce Collider dataclass to pass into _collide() and _collide_pair()
* Replace references to config.friction with per-body friction (Collider.friction)

Fixes #55

Please note:

1. I did some primitive tests within physics_test.py but observed that the simulation is currently unstable with friction due to lateral friction, no matter how large I set the number of substeps. Hence I ended up rolling back, in case someone else can advise. I'd appreciate any help writing tests for these changes in near future.

2. I am preserving the plane-body (and mesh-body?) collision behaviour in _collide(), i.e. only the body's coefficient matters and not the plane's. However, a plane's collision coefficient is pretty important for rolling and sliding motions. I'm not a simulation expert yet, but I'd love to see how e.g. Bullet or MuJoCo implements lateral friction.

### DIFF
--- a/brax/physics/colliders.py
+++ b/brax/physics/colliders.py
@@ -39,6 +39,8 @@ class Collidable:
     self.body = jp.take(body, [body.index[c.name] for c in collidables])
     self.pos = jp.array(
         [vec_to_arr(c.colliders[0].position) for c in collidables])
+    # assuming first collider's material is body's material
+    self.friction = jp.array([c.colliders[0].material.friction for c in collidables])
 
   def position(self, qp: QP) -> jp.ndarray:
     """Returns the collidable's position in world space."""
@@ -65,7 +67,6 @@ class Contact:
     self.vel = vel
     self.normal = normal
     self.penetration = penetration
-
 
 class Cull(abc.ABC):
   """Selects collidable pair candidates for collision detection."""
@@ -145,7 +146,6 @@ class Collider(abc.ABC):
     """
     self.contact_fn = contact_fn
     self.cull = cull
-    self.friction = config.friction
     self.elasticity = config.elasticity
     self.baumgarte_erp = config.baumgarte_erp * config.substeps / config.dt
 
@@ -199,7 +199,7 @@ class OneWayCollider(Collider):
     vel_d = contact.vel - normal_vel * contact.normal
     impulse_d = jp.safe_norm(vel_d) / ((1. / (col.body.mass)) + ang)
     # drag magnitude cannot exceed max friction
-    impulse_d = jp.minimum(impulse_d, self.friction * impulse)
+    impulse_d = jp.minimum(impulse_d, col.friction * impulse)
     dir_d = vel_d / (1e-6 + jp.safe_norm(vel_d))
     dp_d = col.body.impulse(qp, -impulse_d * dir_d, contact.pos)
     # apply collision if penetrating, approaching, and oriented correctly
@@ -257,8 +257,10 @@ class TwoWayCollider(Collider):
     vel_d = contact.vel - normal_vel * contact.normal
     impulse_d = jp.safe_norm(vel_d) / ((1. / col_a.body.mass) +
                                        (1. / col_b.body.mass) + ang)
+    # select friction coefficients (combine by multiplication rule by default)
+    friction = col_a.friction * col_b.friction
     # drag magnitude cannot exceed max friction
-    impulse_d = jp.minimum(impulse_d, self.friction * impulse)
+    impulse_d = jp.minimum(impulse_d, friction * impulse)
     dir_d = vel_d / (1e-6 + jp.safe_norm(vel_d))
     dp_d_a = col_a.body.impulse(qp_a, impulse_d * dir_d, contact.pos)
     dp_d_b = col_a.body.impulse(qp_b, -impulse_d * dir_d, contact.pos)
--- a/brax/physics/config.proto
+++ b/brax/physics/config.proto
@@ -61,6 +61,12 @@ message Collider {
     float size = 1;
     repeated float data = 2;
   }
+  // High-level message for containing elasticity and friction information
+  message Material {
+    // How much surfaces in contact resist translation
+    float friction = 1;  // When unspecified, defaults to global friction coefficient
+    // TODO: add elasticity coefficient (float elasticity = 2);
+  }
   // Position relative to parent body
   Vector3 position = 1;
   // Rotation relative to parent body
@@ -73,6 +79,8 @@ message Collider {
     Capsule capsule = 6;
     HeightMap heightMap = 7;
   }
+  // Material properties (e.g. friction, elasticity) specific to the collider
+  Material material = 8;
 }
 
 // Joints connect two bodies and restrain their movement to a rotational frame.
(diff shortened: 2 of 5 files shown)


namheegordonkim:
@googlebot I fixed it.

brax/physics/colliders.py
```diff
@@ -581,11 +620,13 @@ def _collide_pair(config: config_pb2.Config, body_a: bodies.Body,
...
-  impulse_d = jnp.where(impulse_d < config.friction * impulse, impulse_d,
-                        config.friction * impulse)
+  impulse_d_a = jnp.where(impulse_d < col_a.friction * impulse, impulse_d,
+                        col_a.friction * impulse)
+  impulse_d_b = jnp.where(impulse_d < col_b.friction * impulse, impulse_d,
+                        col_b.friction * impulse)
   dir_d = rel_vel_d / (1e-6 + math.safe_norm(rel_vel_d))
-  dp_d_a = body_a.impulse(qp_a, impulse_d * dir_d, pos_c)
-  dp_d_b = body_b.impulse(qp_b, -impulse_d * dir_d, pos_c)
+  dp_d_a = body_a.impulse(qp_a, impulse_d_a * dir_d, pos_c)
```
erikfrey, line 628: I believe that a single set of material properties governs how two bodies interact.  The way other engines handle this is they let the user specify a policy for how to combine the properties.  See for example Unity:

https://docs.unity3d.com/Manual/class-PhysicMaterial.html

"When two bodies are in contact, the same bounciness and friction effect is applied to both of them..."

So what is missing then is a policy for how to combine the frictions.  I suggest for now we just include a sane default, which is to multiply the frictions together.  So if one material has 0.6, and another has 1.0, then the resulting friction is 0.6.

In some future PR we can perhaps allow the user to specify how to combine (minimum, max, average, multiply, etc)

brax/physics/colliders.py
```diff
@@ -27,6 +27,27 @@
 from jax import ops
 import jax.numpy as jnp
 
+@struct.dataclass
```
erikfrey, line 30: I'm not sure this dataclass is adding much as a grouping concept.  I can't find anywhere in the code where you are referencing `Collider.idx`.  How about just treating friction as a separate scalar for now?

In some future PR we'll clean up these groupings a bit.  They are a bit arbitrarily different at the moment (e.g. CapsuleCapsule has its own struct, while BoxPlane just has a bunch of fields, etc)

erikfrey (changes requested):
Thank you for getting this going!  It's a popular request so I think it'll make some folks happy to get it in.  I proposed a few important design changes in this PR.

After those are addressed, we may need to make a few more relatively minor tweaks, for example updating some of the environment configs to respect the new format.

brax/physics/config.proto
```diff
@@ -65,13 +65,15 @@ message Collider {
...
+  // Friction coefficient
+  optional float friction = 3;
   // A collider may only be one type
   oneof type {
-    Box box = 3;
-    Plane plane = 4;
-    Sphere sphere = 5;
-    Capsule capsule = 6;
-    HeightMap heightMap = 7;
+    Box box = 4;
```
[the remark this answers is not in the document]
namheegordonkim: Good point.

brax/physics/config.proto
```diff
@@ -65,13 +65,15 @@ message Collider {
   Vector3 position = 1;
   // Rotation relative to parent body
   Vector3 rotation = 2;
+  // Friction coefficient
+  optional float friction = 3;
```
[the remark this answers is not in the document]
namheegordonkim: Aha, agreed. I was wondering if there would be a good higher-level container for these properties.

brax/physics/colliders.py
namheegordonkim: Ok, will make a shared friction coefficient using the product rule by default.

brax/physics/colliders.py
namheegordonkim: Yes, separate friction scalar for now would make sense. I was trying to minimize the number of future changes to the signatures of `_collide()` etc., but I think keeping things simple for now would be best.

brax/physics/colliders.py
namheegordonkim: @erikfrey For clarification, should this policy be specific to each body or applied to the entire environment?

namheegordonkim:
Hi @erikfrey , what's the status of this PR?

namheegordonkim:
@erikfrey Thanks for the kind update on the status. Roughly when do you think you can get to merging the changes? My hands are full at the moment, but I might be able to take a crack at it later this coming week.

namheegordonkim:
@erikfrey I got around to tweaking the new code here and there. Take a look and please let me know what you think.

brax/physics/config.proto
```diff
@@ -61,6 +61,12 @@ message Collider {
     float size = 1;
     repeated float data = 2;
   }
+  // High-level message for containing elasticity and friction information
+  message Material {
+    // How much surfaces in contact resist translation
+    optional float friction = 1;  // When unspecified, defaults to global friction coefficient
```
erikfrey, line 67: Let's remove the optional here, as the whole Material is already optional as a Message field.

brax/physics/colliders.py
```diff
@@ -35,10 +35,12 @@ class Collidable:
   corner collidables.
   """
 
-  def __init__(self, collidables: List[config_pb2.Body], body: bodies.Body):
+  def __init__(self, collidables: List[config_pb2.Body], body: bodies.Body, config: config_pb2.Config):
```
erikfrey, line 38: Rather than adding configs to the signatures to fetch global friction, let's just pre-process the config to insert materials with the globals where missing.  We already do this kind of config pre-processing here:

https://github.com/google/brax/blob/main/brax/physics/base.py#L147

So should be straightforward to add more pre-processing.  Then you can just depend on colliders always having materials in the code.

erikfrey (changes requested):
Looks great!  I think we are down to two nits (remove the optional, preprocess the config so everything has a material) - otherwise we are good to go.  Thanks for your patience.    :-)

namheegordonkim:
@erikfrey Thanks for the feedback. Did a number on it.

1. As suggested, I removed `config` from signatures and instead added preprocessing in `base.py` to ensure every collidable object has materials with default friction.
2. I removed the `optional` keyword.

erikfrey:
Code looks perfect.  I suspect the tests are failing as we now have a new policy for combining friction that is different from before.  Grasp test is failing because the ball is slipping out of the hand I suspect.

The fix is probably to update the env configs to account for this, which probably just means that each global friction now has to be sqrt().  to get the old pairwise friction of 0.6, the global friction now needs to be sqrt(0.6).

erikfrey:
@namheegordonkim Thank you!  Merging.

PR #62, opened 2021-09-15
fix integrator, causing incorrect rotational update

erwincoumans:
In a nutshell, the kinetic integrator rotates twice as fast as it should do.
Small check: set time step to 0.01, run 314 steps, the object should rotate half a round (with this patch), but currently it rotates a full round. Once simple fix is replacing the 0.5 factor by 0.25 in line 45 and line 46 in integrators.py. This patch is more efficient though.
In Bullet we use an exponential map with higher accuracy, that can deal with very small angular velocities, read more about it here: https://www.cs.cmu.edu/~spiff/moedit99/expmap.pdf or code here:
https://github.com/bulletphysics/bullet3/blob/master/src/LinearMath/btTransformUtil.h#L32

See attached reproduction case (needs pip3 install meshcat)
[ballplane.zip](https://github.com/google/brax/files/7172910/ballplane.zip)

### DIFF
--- a/brax/physics/integrators.py
+++ b/brax/physics/integrators.py
@@ -40,10 +40,8 @@ def kinetic(_, qp: QP, dt: float, active_pos: jnp.ndarray,
   def op(qp: QP, active_pos: jnp.ndarray,
          active_rot: jnp.ndarray) -> QP:
     pos = qp.pos + qp.vel * dt * active_pos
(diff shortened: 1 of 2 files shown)


cdfreeman-google:
Fixed in an internal merge, and landing externally shortly

PR #75, opened 2021-10-19
Bug fix for infinite recursion when pickling environments.

### LINKED ISSUE #72: Infinite recursion when copying environment
Hello, I stumbled upon this bug today. When I run the following:

```python
from brax.envs import create_fn
import torch
import copy

CUDA_AVAILABLE = torch.cuda.is_available()
if CUDA_AVAILABLE:
    # BUG: (@lebrice): Getting a weird "CUDA error: out of memory" RuntimeError
    # during JIT, which can be "fixed" by first creating a dummy cuda tensor!
    v = torch.ones(1, device="cuda")

env = create_fn('fast')()
env2 = copy.deepcopy(env)
```

I see an infinite recursion error (python 3.9, latest brax master: 73a59dbc598b77667fd65dd623d2eebb87850e52). 

The solution I found was to modify \_\_getattr\_\_ in the Wrapper class from brax/envs/env.py from:

```python
def __getattr__(self, name):
    return getattr(self.env, name)
```
to:
```python
  def __getattr__(self, name):
    if name == "__setstate__":
      raise AttributeError(name)
    return getattr(self.env, name)
```

I found this solution here: https://stackoverflow.com/questions/47299243/recursionerror-when-python-copy-deepcopy, they actually recommend catching any dunder score attribute, I figure the brax team would know better than me if that will cause problems. 

I'd be happy to open a PR with this but it's such a trivial change that I doubt that is worth anyone's time.

sgillen:
Closes #72

### DIFF
--- a/brax/envs/env.py
+++ b/brax/envs/env.py
@@ -94,4 +94,6 @@ class Wrapper(Env):
     return self.env.unwrapped
 
   def __getattr__(self, name):
+    if name == "__setstate__":
+      raise AttributeError(name)
     return getattr(self.env, name)


sgillen:
@googlebot I signed it!

Issue #61, opened 2021-09-14
Applying external forces


namheegordonkim:
Hi,

I think Brax currently doesn't support applying external force/torque to the bodies. I'm getting around this by setting linear and angular velocities with `qp.replace()`. It would be nice to have an explicit function like PyBullet's `applyExternalForce`. Any plan to implement something like this?

cdfreeman-google:
Yes, we do plan to implement this, but haven't had a good concrete need for it yet, and it somewhat complicates the `step` semantics if not done carefully.  I'll follow up here when we get to this (hopefully soon!)

namheegordonkim:
Bumping this request. External forces proved very useful for trajectory optimization in my recent project, and I'd love to use Brax for further development of this. See: https://dl.acm.org/doi/abs/10.1145/3480144

erwincoumans:
Great, closing the issue. If the external force application doesn't work for you, please file a new issue.

PR #99, opened 2021-10-31
External Torque

peabody124:
For discussion:
Beginning to implement external torque inputs. API might make sense to change, though, and is currently incomplete.

The idea is that the frozen field on the Thruster could indicate whether to freeze either force or torque axes, allowing anywhere from 0/1 to 6 DOF for a Thruster.

Currently this ignores the frozen field and provides all possible DOF. In addition, it might be worth adding a different strength field for torque versus force. However, I'm not going crazy in case a preferred approach is an entirely separate class  for external torques than forces.

This builds on #94 and addresses #61


peabody124:
@cdfreeman-google wondering if you have thoughts/comments on the API here

erikfrey:
I REALLY like the choice of using `frozen` to specify DoF - good thinking.  I'll review this PR more in depth in the next day or so, unless @cdfreeman-google gets to it first.

brax/physics/forces.py
```diff
@@ -21,47 +21,56 @@
...
     self.body = jp.take(body, [body.index[f.body] for f in forces])
     self.strength = jp.array([f.strength for f in forces])
-    self.act_index = jp.array(act_index)
+    self.force_index = jp.array(force_index)
+    self.torque_index = jp.array(torque_index)
 
-  def apply_reduced(self, force: jp.ndarray) -> jp.ndarray:
+  def force_apply_reduced(self, force: jp.ndarray) -> jp.ndarray:
     dvel = force * self.strength / self.body.mass
     return dvel, jp.zeros_like(dvel)
```
cdfreeman-google, line 48: Suggestion for slightly simpler interface:  here, either `return dvel` and `return dang` for the torque fn bit, OR use a single function that takes both `force` and `torque` and return the dvel and dang bits together

brax/physics/forces.py
```diff
@@ -74,12 +83,22 @@ def get(config: config_pb2.Config,
...
+    force_dof = int(sum(pos_mask))
+    force_act_index = tuple(range(current_index, current_index + force_dof))
+    force_indices.append(force_act_index)
+    current_index += force_dof
+
+    rot_mask = 1. * jp.logical_not(vec_to_arr(force.frozen.rotation))
+    torque_dof = int(sum(rot_mask))
+    torque_act_index = tuple(range(current_index, current_index + torque_dof))
+    torque_indices.append(torque_act_index)
+    current_index += torque_dof
```
cdfreeman-google, line 99: Make sure you add `frozen: all` fields to the rotation fields on the forces in the inverted_pendulum and inverted_double_pendulum environments

cdfreeman-google:
Agree with Erik, this is great stuff.  Thanks so much for the proposal!  Added one interface comment, and pointed out the one other place in the codebase that I think this touches currently.

brax/physics/forces.py
erikfrey: +1 - we have other parts of the code that only return dvel or dang, so not breaking any conventions by doing this.

erikfrey: But actually I slightly prefer the second option: just have one function.  Also, don't forget to fix the return type!

```python
def apply_reduced(self, force: jp.ndarray, torque: jp.ndarray) -> Tuple[jp.ndarray, jp.ndarray]:
  ...
```

brax/physics/forces.py
erikfrey: +1

erikfrey (changes requested):
Agree with Daniel on comments, I think this is ready barring the two nits.

peabody124:
@erikfrey @cdfreeman-google thanks for the positive responses! It's still not fully functional as it doesn't properly respect the frozen fields yet. I'm working on this still. I was running into issues with handling a flexible number of channels (particularly in the case of multiple forces). I've done a bit more learning about jax and think I have a valid approach by vectorizing over both a mask array and an index array (padded with zeros for masked out dimensions). I hope to have this updated by the weekend with comparisons on the pendulum channels to ensure no regression.

p.s. there is already a bit of a fudge in that the pendulum classes override action_size to be 1 despite the force using 3 elements. I'll hopefully have it freeze all but one axis and be able to remove that class definition.

brax/physics/forces.py
peabody124: @erikfrey do you still prefer the second option now that I added the mask field? It would make the convention quite clunky having all four inputs. For now I went with @cdfreeman-google 's suggestion of only returning what it needs. That also seems like a slightly cleaner abstraction since the calculation are entirely independent.

brax/physics/forces.py
peabody124: Added ```frozen { position { x: 0.0 y: 1.0 z: 1.0} rotation { x: 1.0 y: 1.0 z: 1.0 } ``` to both so only the needed axis works. I confirmed both environments produce stable behave and training curves look similar.

Side question. There is no convention to say ```frozen { rotation : all }``` that I can see

peabody124:
@erikfrey I believe all nits should be handled.

erikfrey:
Hi @peabody124, this is a tricky design decision - in general I like the idea of moving to generalized forces that can take a variable number of the 6dofs, and we will in a future version of Brax in which we do reduced coordinates (see #104 )

Looking at this closely now, I can see that it works but is a bit out of step with how we construct our other helper classes.  Joint has Revolute, Universal, Spherical, which each purposefully handle distinct shapes so that we vmap over them individually.  In the same way Actuator devolves under the hood into an Actuator1D, Actuator2D, Actuator3D classes, so that we can vmap over them.

So I think for now, we should stick to that pattern.  In the future when we move to more generalized joint/actuator system we will rip off the bandaid and change all of this at once.  So concretely I think that means: leave Thruster as a dumb/simple class that for now is hard coded to take in 3 positional dimensions, and add a new message type and class called, say, Twister that is also dumb/simple and takes in 3 angular dimensions.

I realize this is a big left turn from what you have so far and apologize for that - Daniel and I should have looked at this more closely early on when you first proposed it and anticipated this design issue.  If you'd like to update this same PR, we'll get this in!

peabody124:
Sure. I can split this into a separate Twister. I don't mind stripping out the action dimension selection too. It's actually not needed for the purpose I'm currently using the residual forces for.

That being said, before doing so I'd just like to play devil's advocate for this implementation with another use case I'd been considering. Say you want to make a quadcopter (semi accurately) then you would want four thrusters with force along the Z axis only and torque around the Z axis only. A 3D twister/thruster format would prohibit making that type of model without an awkward environment that maps the action space onto the system's action dimensions.

The current implementation, by using the additional (precomputed) mask helper variable, still does allow vmap-ing along them.

Anyway, if you want to split into separate Thruster class that is always 3 axes, I can do that.

erikfrey:
You are 100% correct that with the proposed design, you'd have to do something like:

```python
def step(self, state: env.State, action: jp.ndarray) -> env.State:
  # quadcopter thrusters only work in +/- Z
  act = jp.array([0, 0, action[0], 0, 0, action[1], 0, 0, action[2], 0, 0, action[3]]) 
  qp, info = self.sys.step(state.qp, act)
  ...
```

Which is less powerful than what you have built.  Normally I'd agree with your devil's advocate position, but we are going to soon start working #104 which is going to involve some pretty big changes both under the hood and to the config format.

In this case I'm prioritizing uniformity of the underlying implementation over expressivity - I'd prefer to keep things simple for now to ease the pain of that upcoming refactor.

erikfrey:
External torque is now implemented in 3a197dba959901fab16333ccde7e44e8f1981cfd - please try it out and let us know if there are any issues.

peabody124:
@erikfrey is there still external torque in v2? I was working on porting some code to that but I'm not seeing they still exist...

btaba:
Hi @peabody124 , we don't have external torque in v2, but if you'd like to implement it, I can provide some pointers

peabody124:
@btaba that would be great. happy to work on PR since I'd love to update my code soon

btaba:
Hey @peabody124 

Actuator sites are more likely to be first implemented in MJX (https://github.com/google/brax/discussions/409) following MuJoCo's implementation. Here's where to start:

https://github.com/google-deepmind/mujoco/blob/e57b34303a13195bf5fe87d37cb778321130a229/src/engine/engine_core_smooth.c#L1001

I'm assuming you're using the `generalized` backend. If not you can follow the implementation in v1 https://github.com/google/brax/blob/main/brax/v1/physics/forces.py

Issue #89, opened 2021-10-28
Make accelerator-friendly versions of a subset of the gym wrappers


erikfrey:
Candidates are:

gym.wrappers.RecordEpisodeStatistics
gym.wrappers.ClipAction
gym.wrappers.NormalizeObservation
gym.wrappers.TransformObservation
gym.wrappers.NormalizeReward
gym.wrappers.TransformReward

And also possibly a new wrapper:

gym.wrappers.RecordHTML

/cc @vwxyzjn @jkterry1

jkterry1:
So I had a discussion with Costa, and I don't know what the best way to proceed forward here is.

Right now, literally no current Gym wrappers can be used with Brax environments because they're all numpy based which has almost no interoperability. This would mean that, for full support (Phys2D, Phys3D, classic control, etc.), there would need to be duplicates of all the Gym wrappers written in Jax.

This is rather undesirable for several reasons:
-Copying and maintaining copied code sucks
-gym.wrappers needs a big overhaul, doesn't have various key wrappers, certain wrappers are missing key features, and so on, so there's not a reference set to copy
-We'd have to make sure that the Jax and numpy gym wrappers are have identical function for reproducibility papers when people are describing wrappers in the literature
-There isn't a clear place for the Jax wrappers to even live if classic control is jax-ified

I know you wrote a conversion code between Jax and torch tensors to get things to work, is there any way to shim conversion code to get Gym wrappers to work as is to avoid this whole situation?



An additional related point is- where should the Jax->torch (and possible other) Gym adapters live, given things like the plan for Jax based classic control? 

One reasonable candidate location is https://github.com/Farama-Foundation/SuperSuit. SuperSuit is a collection of Gym and PettingZoo wrappers, once gym.wrappers is complete and stable those will be removed, and the pettingzoo wrappers will be merged into pettingzoo.wrappers, like Gym. However, SuperSuit is still going to retain various preprocessing wrappers that don't make sense to house in either Gym or PettingZoo, such as ones purely for manipulating Gym or PettingZoo vector environments or converting the PettingZoo environments to Gym environments (this is something you actually want to do in certain odd cases). It doesn't have to go there is there's a better option or one that you'd prefer or something, it would just be odd to use a Brax wrapper for classic control for instance.

erikfrey:
@jkterry1 please take a look at this shim we put in Brax a while back:

https://github.com/google/brax/blob/main/brax/jumpy.py

[...]

PR #113, opened 2021-11-11
Normalization Wrappers

DavidSlayback:
Partial #89 

This is an initial attempt at doing performant normalization wrappers akin to the standard gym normalization wrappers. This includes NormalizeObservation and NormalizeReward, using the same approach as Gym/StableBaselines.

I've implemented it at the brax_env.env level instead of the gym level and tested that it's compatible with JIT and VectorWrapper environments.

Open to ideas of how to make this better...I'm not a fan of the hack I did to include with running statistics in the environment state while keeping them out of vmap logic! But I wanted to get these started :)

### DIFF
--- a/brax/envs/wrappers.py
+++ b/brax/envs/wrappers.py
@@ -14,7 +14,7 @@
 
 """Wrappers for Brax and Gym env."""
 
-from typing import ClassVar, Optional
+from typing import ClassVar, Optional, NamedTuple
 
 from brax import jumpy as jp
 from brax.envs import env as brax_env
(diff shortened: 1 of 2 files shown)


brax/envs/wrappers.py
```diff
@@ -97,6 +98,84 @@ def where_done(x, y):
...
+    if 'running_obs' not in state.info:
+      obs_like = state.obs[0] if hasattr(self.env, 'batch_size') else state.obs
+      # obs_like = jp.atleast_2d(obs_like)[0]
+      state.info['running_obs'] = RunningMeanStdState(jp.zeros_like(obs_like), jp.ones_like(obs_like))
+    state.info.update(running_obs=update_running_mean_std(state.info['running_obs'], state.obs))
+    return state.replace(obs=normalize_with_rmstd(state.obs, state.info['running_obs']))
+
+
+  def step(self, state: brax_env.State, action: jp.ndarray) -> brax_env.State:
+    running_obs = state.info.pop('running_obs')  # hack to avoid vmap :(
```
erikfrey, line 147: I think this is an OK solution for now.

brax/envs/wrappers.py
```diff
@@ -97,6 +98,84 @@ def where_done(x, y):
...
+class RunningMeanStdState(NamedTuple):
+  """Running statistics for observtations/rewards"""
+  mean: jp.ndarray
+  var: jp.ndarray
+  count: float = 1e-4
+
+
+def update_running_mean_std(std_state: RunningMeanStdState, batch: jp.ndarray) -> RunningMeanStdState:
+  """Update running statistics with batch of obsrvations (Welford's algorithm)"""
+  batch = jp.atleast_2d(batch)[0]  # Account for unbatched environments
```
erikfrey, line 110: Did you forget to add some changes to jumpy for this to work?

erikfrey (changes requested):
Hi David - thank you for opening this PR!  Please see #89 for ongoing discussion.  This is an area that is still quite in flux.  For what it's worth, I would like to see most of these env wrappers moved out of Brax and under the purview of gym, so I'm hesitant to make any big changes in the Brax repo until that is all sorted out.

In particular, things like observation normalization are handled in the brax learners outside of the environments, and that's a deliberate design decision.  The main reason being that we often run on multiple devices and want to be careful to aggregate running statistics across device, which requires special handling in the trainer.

So I think I defer to @vwxyzjn and @jkterry1 with what to do with this particular PR.

That said!  I'd be happy to accept a minimal PR here just to get in a tight version of observation normalization, so that we could simplify (and possibly improve the perf a tiny bit) of this pytorch training colab: https://colab.sandbox.google.com/github/google/brax/blob/main/notebooks/training_torch.ipynb

If you'd like to have a go at making just an observation normalization wrapper, and then updating the colab to show it working, I'd be happy to shephered that in while the grander gym planning is underway.

Issue #114, opened 2021-11-13
HTML Rendering Bug/Undesired Behavior in JupyterLab


sgillen:
Hello, I've been experiencing issues when making multiple calls to the html renderer while using jupyterlab (3.1.18, under both chrome and firefox).

I can produce the behavior by running the following code in my instance of jupyter lab (with each #%% representing a new cell)

```python

#%% Cell 0
from IPython.display import HTML

import brax
from brax import envs
from brax import jumpy as jp
from brax.io import html

#%% Cell 1 

env_name = "ant" 
env_fn = envs.create_fn(env_name=env_name)
env = env_fn()
state = env.reset(rng=jp.random_prngkey(seed=0))

HTML(html.render(env.sys, [state.qp])) # First call, works fine

#%% Cell 2

HTML(html.render(env.sys, [state.qp])) # Second call, creates a blank space under cell 2, and silently adds a new underneath the viewer created when running cell 1 (which is invisible without scrolling).
```

As the comment explains, every call after the first results in the new viewer being added underneath the first cell. If that's confusing I've attached a screenshot that explains what I'm seeing. Even worse, if I open a new tab/notebook and attempt to render there, it will add a third viewer to the first cell of the original tab!

This differs from the behavior seen in colab notebooks, where each call to html.render makes a viewer in the cell it's called in (which is the desired behavior here, for me). I believe the issue is conflicting div ids for the viewer windows, perhaps colabs are doing something to give unique ids for html generated in different cells? 

Either way I have a fix that works for me here: (https://github.com/google/brax/pull/115) which adds a uuid to the viewer name.

![brax_render_issue](https://user-images.githubusercontent.com/9152639/141659519-6ae9f4fd-1e15-47c5-b0ce-9f23b9862660.png)

erikfrey:
Hi Sean,

Did you try an IFrame?

```python

from IPython.display import IFrame

...

IFrame(src=html.render(env.sys, [state.qp])) # First call, works fine
```

I see your PR and agree that making the viewer id unique fixes the immediate problem, but it may cause other problems down the line to allow use cases where multiple HTML files (with opening/closing tags for html and body, multiple script includes, etc) are appended into the same html file.

You're right that in colab this is different!  I think we are taking for granted that in Colab, outputs are automatically injected into an IFrame, I believe.

sgillen:
Hmm I can't get that to work, does that work for you?. Modifying your idea slightly, I can get something that almost works:
```python
from IPython.display import IFrame

...

[...]

PR #115, opened 2021-11-13
fix html.render so it makes a new viewer window for every call in jupyterlab

sgillen:
Would close https://github.com/google/brax/issues/114.
I've verified that this does not change the behavior when running colabs, and replicates the colab behavior in jupyter, and that the new render code still respects the height parameter passed in. But perhaps there is some use case I am not thinking of that this will break.

### DIFF
--- a/brax/io/html.py
+++ b/brax/io/html.py
@@ -16,6 +16,7 @@
 
 import json
 import os
+import uuid
 from typing import List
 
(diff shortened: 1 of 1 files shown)


erikfrey:
Closing for now, but happy to revisit if the issue comes up again for other users.

PR #126, opened 2021-11-26
Replace dat.gui import with lil.gui

### LINKED ISSUE #125: Replace Dat GUI Dependency
THREE.js has retired Dat GUI officially, as per this PR: https://github.com/mrdoob/three.js/pull/22765

As of today (November 26, 2021), HTML rendering breaks due to missing import.

`viewer.js` should reflect this change by changing the import line as follows:

change

```
import {GUI} from 'https://threejs.org/examples/jsm/libs/dat.gui.module.js';
```

to

```
import {GUI} from 'https://threejs.org/examples/jsm/libs/lil-gui.module.min.js';
```

namheegordonkim:
Fixes #125

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -5,7 +5,7 @@
 
 import * as THREE from 'https://threejs.org/build/three.module.js';
 import {OrbitControls} from 'https://threejs.org/examples/jsm/controls/OrbitControls.js';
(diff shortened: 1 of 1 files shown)


erikfrey:
Good catch - thanks for fixing this!

PR #134, opened 2021-12-11
fix error in createHeightMap function of system.js

chaihahaha:
THREE.ParametricGeometry has been moved to /examples/jsm/geometries/ParametricGeometry.js


chaihahaha:
I've signed CLA, please rescan.

erikfrey:
@chaihahaha Thanks for finding this!  I'd love to merge it but it still says "Missing CLA" - I've seen this before and it's usually one of the issues here: https://github.com/google/brax/pull/134/checks?check_run_id=4493113189

Can you confirm if that all looks correct and then we can try again?

chaihahaha:
I've corrected the email in my CLA, please try again.

PR #138, opened 2021-12-14
fix error in createHeightMap function of system.js

chaihahaha:
THREE.ParametricGeometry has been moved to /examples/jsm/geometries/ParametricGeometry.js


erikfrey:
Thank you!

PR #145, opened 2022-01-06
Changing humanoid termination z-threshold

lerrytang:
Hi,

I've changed the z-threshold for termination to prevent the agent from moving on knees.
The change does not affect the released notebooks (e.g., "Brax Training").


erikfrey:
Thank you!

erikfrey:
Oops, actually, we have some internal tests that are sensitive to this change.  I'm going to temporarily revert it and we'll get this in next week.  Thanks!

PR #153, opened 2022-01-20
Support variance in `jumpy`

vwxyzjn:


jumpy/jumpy.py
```diff
@@ -164,6 +164,9 @@ def mean(a: ndarray, axis: Optional[int] = None) -> ndarray:
   """Compute the arithmetic mean along the specified axis."""
   return _which_np(a).mean(a, axis=axis)
 
+def var(a: ndarray, axis: Optional[int] = None) -> ndarray:
+  """Compute the arithmetic mean along the specified axis."""
```
erikfrey, line 168: Please update the docstring!

erikfrey (changes requested):
One nit, otherwise LGTM

jumpy/jumpy.py
vwxyzjn: Done

PR #158, opened 2022-02-01
Add Acrobot Environment, Remove Redudant Inverted Double Pendulum Observations

sgillen:
Hello,

I've been experimenting with some swing-up pendulum environments, I thought Brax might find some of them useful / interesting.

I am proposing adding an Acrobot environment to Brax. The one I am submitting here is a bit different and more difficult than the one in I.E. gym, since the action state in continuous, and to "solve" the environment an agent must both swing up and balance the system. I like this environment because it is deceptively difficult for most model free RL  with it (see for example [this paper](https://arxiv.org/pdf/2012.11663.pdf) I wrote a couple of years ago).

I've got my own twist on APG that works well for this acrobot, but haven't yet been able to get good performance on it from any of the brax algorithms (including APG). I would be very interested to hear if anyone gets the brax RL working well with the environment!

In this PR I also remove some redundant / useless observations from the double inverted pendulum. 

Let me know what you all think!

Edit: Also, [here is](https://colab.research.google.com/drive/1cZcaQtRjfM91Iot1cBZhITLti5FmQ0mW?usp=sharing) a basic smoke test training set up in collab.

### DIFF
--- a/brax/envs/__init__.py
+++ b/brax/envs/__init__.py
@@ -38,6 +38,8 @@ from brax.envs import wrappers
 from brax.envs.env import Env, State, Wrapper
 import gym
 
+from brax.envs import acrobot
+
 
 _envs = {
     'ant': ant.Ant,
@@ -54,6 +56,7 @@ _envs = {
     'reacherangle': reacherangle.ReacherAngle,
     'swimmer': swimmer.Swimmer,
     'ur5e': ur5e.Ur5e,
+    'acrobot': acrobot.Acrobot,
     'walker2d': walker2d.Walker2d,
 }
 
--- a/brax/envs/acrobot.py
+++ b/brax/envs/acrobot.py
@@ -60,14 +60,15 @@ class Acrobot(env.Env):
   def step(self, state: env.State, action: jp.ndarray) -> env.State:
     """Run one timestep of the environment's dynamics."""
     qp, info = self.sys.step(state.qp, action)
+    (joint_angle,), (joint_vel,) = self.sys.joints[0].angle_vel(qp)
     obs = self._get_obs(qp, info, joint_angle, joint_vel)
 
     alive_bonus = 10.0
-    dist_penalty = obs[0]**2 + obs[1]**2
-    vel_penalty = 1e-3*(obs[2]**2 + obs[3]**2)
+    dist_penalty = joint_angele[0]**2 + joint_angle[1]**2
+    vel_penalty = 1e-3*(joint_bel[0]**2 + joint_vel[1]**2)
     r =  alive_bonus - dist_penalty - vel_penalty
-
     done = jp.float32(0);
+    
     state.metrics.update(
         dist_penalty=dist_penalty,
         vel_penalty=vel_penalty,
@@ -82,8 +83,6 @@ class Acrobot(env.Env):
   def _get_obs(self, qp: brax.QP, info: brax.Info, joint_angle: jp.ndarray,
                joint_vel: jp.ndarray) -> jp.ndarray:
     """Observe acrobot body position and velocities."""
-
-    (joint_angle,), (joint_vel,) = self.sys.joints[0].angle_vel(qp)
  
     return jp.concatenate((joint_angle, joint_vel))
 
--- a/brax/envs/acrobot.py
+++ b/brax/envs/acrobot.py
@@ -1,4 +1,4 @@
-# Copyright 2022 The Brax Authors.
+# Copyright 2021 The Brax Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


erikfrey:
Nice!  Tagging in @cdfreeman-google to review this, but at a high lvel I'm all for an acrobot environment.  Have you compared it to an analytical acrobot such as [this one in deluca](https://github.com/google/deluca/blob/main/deluca/envs/classic/_acrobot.py)?

I'm sure @cdfreeman-google would also love to hear details about your twist on APG.

cdfreeman-google:
Ooh, very curious to know what tricks you added to APG to get this to work.  I did a little bit of hparam sweeping with ppo and couldn't get it to train.

sgillen:
Sure @cdfreeman-google , so the version of "apg" I am working on lives here. https://github.com/sgillen/apg/blob/master/ce_apg.py

I was inspired by the Parallel Basin Hopping approach used in the TDS paper. The basic idea is to use APG as a local optimization method, and combine it with an outer loop global optimizer.  

So in pseudo code: 

```
Initialize a single policy P.
while not_converged:
    Generate N initial policies from P by adding random noise. 
    Run N instances of APG in parallel from the N initial policies, and collect the N trained policies and their expected reward.
    Use some update rule to update policy P from the N trained policies.   
```

In PBH the update rule is basically just picking the best of the N policies as the new policy (there's a bit more to it than that but that's the idea). 

I found that using a cross entropy method for updates worked better for me, you could also use for example ARS with some tweaking (though I've yet to try that). So I've tentatively been calling my algorithm  "cross entropy analytic policy gradients" (CE-APG)

However there are some other key differences between the version of APG I made for myself and the one currently in brax. 
- I found using GRUs instead of a normal MLP let me get away with no truncating gradients (at least in environments without contact). 
- I've found the gradients are much faster when using CPU, so my algorithm is designed to work on CPU, but perhaps it can be made even faster by leveraging TPUs, not sure. 
- I am also using deterministic polices, rather than parameterize  and sample from a probability distribution to get the current action. I thought that this would speed up the training, but it could be that sampling actions actually smooths things out and lowers the variance of the gradient?  (Btw I read your recent preprint: https://arxiv.org/abs/2111.05803 very interesting work!). I've yet to test out stochastic policies for my algorithm. 

I've done some ablation studies, so far focusing on the acrobat and swingup versions of the pendulum envs. APG and CEM by themselves with the GRU and deterministic polices still performs poorly, as does CE-APG without the GRU. As I've said I've yet to test if the deterministic vs stochastic policies make a big difference. 

I'd love to get this working with environments with contact (that was the original goal), I usually just get NaNs though. Trying to get to the bottom of that now. 

Anyway I was trying to keep that short but it got pretty long anyway. I'm happy to go into more detail though if you are interested!

sgillen:
And @erikfrey I haven't compared it to Deluca yet, but I can do so tomorrow.

sgillen:
Ok, so the Deluca environment seems to be a copy of the OpenAI version I talked about above, nothing wrong with that but it makes comparisons not super meaningful (the simulation dt for example is 200ms!!).

I compared to my own [acrobot](https://github.com/sgillen/seagul/blob/master/seagul/envs/classic_control/acrobot.py) (the  dynamics come from the derivation [here](https://www.sciencedirect.com/science/article/pii/S1474667017474040)), using the same intertial/simulation parameters as the brax env I am submitting, qualitatively the behavior is the same, see in this [collab](https://colab.research.google.com/drive/1Cjg1NMOygcITC5rfI4_l5ywgLXuqpgiR?usp=sharing). I don't expect the two simulations to exactly match, due to brax's damping, the way joints constraints are handled etc.

Also, not sure if this is worth noting, when using the inertial params from the original acrobot paper, brax became unstable, to fix this I upped the mass for each link (from 1kg -> 2.5kg). The version I am submitting has the heavier 2.5kg links.

cdfreeman-google:
Very cool! Some comments/questions below

> I found using GRUs instead of a normal MLP let me get away with no truncating gradients (at least in environments without contact).
Interesting! This is a little bit surprising to me.  I wonder if this would make optimizing Ant work without truncations...

> I've found the gradients are much faster when using CPU, so my algorithm is designed to work on CPU, but perhaps it can be made even faster by leveraging TPUs, not sure.
Are you hitting a bottleneck with JIT time, or the actual serial execution time is choking on TPU?  Having the gradient does obviate the need for estimating a gradient with a large batch, but I suspect you could massively parallelize that outer basin-hopping optimization, too, if you were using a TPU.

>I am also using deterministic polices, rather than parameterize and sample from a probability distribution to get the current action. I thought that this would speed up the training, but it could be that sampling actions actually smooths things out and lowers the variance of the gradient? (Btw I read your recent preprint: https://arxiv.org/abs/2111.05803 very interesting work!). I've yet to test out stochastic policies for my algorithm.

Yeah this seems fine.  I'd be surprised if there was a dramatic speed difference, but it's definitely extra code complexity to keep around (e.g., plumbing randomness for the policy all over the place).

> Also, not sure if this is worth noting, when using the inertial params from the original acrobot paper, brax became unstable, to fix this I upped the mass for each link (from 1kg -> 2.5kg). The version I am submitting has the heavier 2.5kg links.

Hmm, could you try with double the substeps?  I'm personally fine with upping the mass, but it's nice to know when / why instabilities are arising.

sgillen:
Yes I was also wondering about using the GRU/LSTM on the "vanilla" APG for the Ant. It might require some serious changes to the plumbing to support recurrent policies everywhere though, not sure if that's on the TODO list anyway. As I've said I keep getting NaNs when trying my approach on Ant, let me know if you have any tips for avoiding those (I haven't had much time to play with this recently).

The bottle neck I'm finding is the the actual execution of the local APG search, though do note I've not tested much on TPU, just on my own GPU. 

With double the substeps and mass = 1.0 things are in fact stable. I sort of prefer the heavier version just because it's a bit faster and the resulting motions tend to look a bit nicer, but I don't think it's a big deal either way.

erikfrey:
Changes look great.  Happy to keep the mass as is in the PR.  Thanks @sgillen !

PR #159, opened 2022-02-02
Add checkpoint_dir to PPO

susumuota:
I added `checkpoint_dir` option to `ppo.py`. Also, I tweaked `sac.py` for consistency.

Thanks you for the great physics engine!


erikfrey:
Very nice change - thank you!

PR #179, opened 2022-03-18
Additional jumpy functions

DavidSlayback:
- Control flow:
  - While loop
  - Cond
- Random:
  - randint
  - choice
- Miscellaneous
  - atleast_1/2/3d
  - meshgrid

Following up from #174, these are the additional jumpy functions I wound up implementing for my own brax environments.

PR #190, opened 2022-04-21
Remove local jumpy repo for Farma-Foundation jumpy

pseudo-rnd-thoughts:
This PR removes the local jumpy repository in favour of the [Farma-Foundation jumpy repository](https://github.com/Farama-Foundation/Jumpy)
The advantage of this is that jumpy is planning on being used by [openai/gym](https://github.com/openai/gym) (and hopefully more projects) and this allows all users to share a single jumpy implementation

The two projects at currently identical - we have not merged https://github.com/google/brax/pull/179 at this time

@jkterry1


erikfrey:
@pseudo-rnd-thoughts thanks for opening this PR.  I think for the time being it makes sense to keep two jumpies going.  Here are my thoughts on pros/cons:

We're very happy for Gym to use jumpy if it solves the same problems it did for us: avoiding jax jit latency for lightweight interactive simulation.  There's possibly even some future world where jumpy is a minimal language between jax, pytorch, numpy, warp, who knows...?

BUT:

Jumpy is getting new functions from release to release, and Google's mechanism for releasing code to github is a bit clunky.  It will be a fair bit trickier for us to push Brax changes to github if it includes changes to jumpy.  The flow would be something like:
  - manually copy our internal jumpy changes to a jumpy PR
  - open the PR, go through review, get it merged
  - version bump jumpy and push to pypi
  - update the version req in brax, then push to github

vs what we do today:
  - just push to github

These problems will go away once the jumpy API is stable, because then we can rely on jumpy not changing from one push to the next.  We also know it's okay for two jumpys to live in two separate namespaces, so I don't think if a Gym user uses jumpy, and brax uses brax.jumpy, there will be no problems.

So my suggestion is to wait for the dust to settle a bit on this transition, and after jumpy is stable we can make this kind of cleanup.  Does that make sense?

pseudo-rnd-thoughts:
Hi @erikfrey, thanks for the update. That makes sense for us at Gym and I understand the problem for you 
I will leave this PR open for when jumpy becomes stable and the project can move to the farama-foundation's version
We will continue to develop jumpy for the requirements of gym and try to keep our version of jumpy update with yours to make the two versions compatible with each other  
I hope this plan works for both of us

Issue #198, opened 2022-05-26
[Question] How to debug training algorithms?


vuoristo:
Hi,

I'm trying to debug the PPO algorithm from `brax.training.ppo` while disabling jax jitting. This causes an error with jumpy trying to call a numpy method a Traced jax array. See the command line and log below. Is there a way to run this without jit turned on?

```bash
$ JAX_DISABLE_JIT=1 python -m brax.training.learner
I0526 20:02:54.884144 140223632732992 metrics.py:42] Hyperparameters: {'log_frequency': 10, 'num_envs': 4, 'total_env_steps': 50000000}
I0526 20:02:54.966016 140223632732992 xla_bridge.py:330] Unable to initialize backend 'tpu_driver': NOT_FOUND: Unable to find driver in registry given worker:
I0526 20:02:57.443377 140223632732992 xla_bridge.py:330] Unable to initialize backend 'tpu': INVALID_ARGUMENT: TpuPlatform is not available.
I0526 20:02:57.444162 140223632732992 ppo.py:221] Device count: 8, process count: 1 (id 0), local device count: 8, devices to be used count: 8
Traceback (most recent call last):
  File "/data/tucana/risrio/brax/brax/training/learner.py", line 229, in <module>
    app.run(main)
  File "/data/tucana/risrio/miniconda3/envs/braxbox/lib/python3.9/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/data/tucana/risrio/miniconda3/envs/braxbox/lib/python3.9/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "/data/tucana/risrio/brax/brax/training/learner.py", line 163, in main
    inference_fn, params, _ = ppo.train(
  File "/data/tucana/risrio/brax/brax/training/ppo.py", line 243, in train
    first_state = reset_fn(key_envs)
  File "/data/tucana/risrio/miniconda3/envs/braxbox/lib/python3.9/site-packages/jax/_src/traceback_util.py", line 162, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/data/tucana/risrio/miniconda3/envs/braxbox/lib/python3.9/site-packages/jax/_src/api.py", line 1485, in vmap_f
    out_flat = batching.batch(
  File "/data/tucana/risrio/miniconda3/envs/braxbox/lib/python3.9/site-packages/jax/linear_util.py", line 168, in call_wrapped
[...]

PR #199, opened 2022-05-28
Add an environment variable to force jumpy to use jnp

vuoristo:
Helps with #198. These changes enable forcing jumpy to use jnp via an environment variable. This is needed when jumpy is used with jax tracing but with jit explicitly disabled.

In order to make the flag work, the `jp._which_np` has to be changed so that it checks for nested inputs. It's not clear to me what the purpose of the original `and not isinstance(a, onp.ndarray)` condition is.

Changing the jax branch behavior of `jp.index_update` was required to make this code work with `brax.training.ppo`.

There are quite a few distinct cases where jumpy needs to pick between jnp and onp. I checked that this runs with `brax.training.ppo` with jit enabled or disabled but there could be other cases that running that code didn't hit.

### DIFF
--- a/brax/jumpy.py
+++ b/brax/jumpy.py
@@ -17,7 +17,6 @@
 
 from typing import Any, Callable, List, Optional, Sequence, Tuple, TypeVar, Union
 import builtins
-import os
 
 import jax
 from jax import core
@@ -32,11 +31,10 @@ inf = onp.inf
 float32 = onp.float32
 int32 = onp.int32
 
-FORCE_JNP = os.getenv("JUMPY_FORCE_JNP", 'False').lower() in ('true', '1', 't')
 
 def _in_jit() -> bool:
-  """Returns true if currently inside a jax.jit call."""
-  if FORCE_JNP:
+  """Returns true if currently inside a jax.jit call or jit is disabled."""
+  if jax.config.jax_disable_jit:
     return True
   return core.cur_sublevel().level > 0
 
--- a/brax/jumpy.py
+++ b/brax/jumpy.py
@@ -16,12 +16,14 @@
 """Numpy backend for JAX that is called for non-jit/non-jax arrays."""
 
 from typing import Any, Callable, List, Optional, Sequence, Tuple, TypeVar, Union
+import builtins
 
 import jax
 from jax import core
 from jax import custom_jvp
 from jax import numpy as jnp
 import numpy as onp
+import tree
 
 ndarray = Union[onp.ndarray, jnp.ndarray]  # pylint:disable=invalid-name
 tree_map = jax.tree_map  # works great with jax or numpy as-is
@@ -37,10 +39,11 @@ def _in_jit() -> bool:
 
 
 def _which_np(*args):
-  """Returns np or jnp depending on args."""
-  for a in args:
-    if isinstance(a, jnp.ndarray) and not isinstance(a, onp.ndarray):
-      return jnp
+  checker = lambda a: (
+    isinstance(a, (jnp.ndarray, jax.interpreters.batching.BatchTracer))
+    and not isinstance(a, onp.ndarray))
+  if builtins.any(tree.flatten(tree.map_structure(checker, args))):
+    return jnp
   return onp
 
 
--- a/brax/jumpy.py
+++ b/brax/jumpy.py
@@ -24,7 +24,6 @@ from jax import core
 from jax import custom_jvp
 from jax import numpy as jnp
 import numpy as onp
-import tree
 
 ndarray = Union[onp.ndarray, jnp.ndarray]  # pylint:disable=invalid-name
 tree_map = jax.tree_map  # works great with jax or numpy as-is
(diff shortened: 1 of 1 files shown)


brax/jumpy.py
```diff
@@ -30,17 +33,21 @@
 float32 = onp.float32
 int32 = onp.int32
 
+FORCE_JNP = os.getenv("JUMPY_FORCE_JNP", 'False').lower() in ('true', '1', 't')
```
erikfrey, line 36: just to future-proof this (jumpy may handle also one day handle, say, torch tensors... who knows!), can we make this:

JUMPY_FORCE_LIB=JAX|NUMPY

and it forces jnp if jax, forces np if numpy?

brax/jumpy.py
```diff
@@ -30,17 +33,21 @@
...
 
 def _which_np(*args):
-  """Returns np or jnp depending on args."""
-  for a in args:
-    if isinstance(a, jnp.ndarray) and not isinstance(a, onp.ndarray):
-      return jnp
+  checker = lambda a: (
+    isinstance(a, (jnp.ndarray, jax.interpreters.batching.BatchTracer))
+    and not isinstance(a, onp.ndarray))
+  if builtins.any(tree.flatten(tree.map_structure(checker, args))):
```
erikfrey, line 49: can you use jax.tree_map here instead?  that way we can avoid importing `tree`

it's possible that brax already requires `tree` via some other dependency, but i'd prefer to avoid importing it, as then we'd have to add it to the setup.py requirements.

brax/jumpy.py
```diff
@@ -30,17 +33,21 @@
...
+    return True
   return core.cur_sublevel().level > 0
 
 
 def _which_np(*args):
-  """Returns np or jnp depending on args."""
-  for a in args:
-    if isinstance(a, jnp.ndarray) and not isinstance(a, onp.ndarray):
-      return jnp
+  checker = lambda a: (
```
erikfrey, line 46: interesting - would you want to support FORCE_JNP here too?  or not?

brax/jumpy.py
```diff
@@ -30,17 +33,21 @@
...
+    isinstance(a, (jnp.ndarray, jax.interpreters.batching.BatchTracer))
```
erikfrey, line 47: fyi (since you asked), back in jax v0.2.17, isinstance(my_jax_array, numpy.ndarray) would return True: https://github.com/google/jax/blob/jax-v0.2.17/jax/_src/numpy/lax_numpy.py#L93

so to support older versions of jax we had to add the check `not isinstance(a, onp.ndarray)`

erikfrey (changes requested):
Thank you for doing this!  Overall looks great, just a few questions/comments

brax/jumpy.py
vuoristo: I see, that makes sense.

brax/jumpy.py
vuoristo: Makes sense. I pushed a change that uses `jax.tree_map` and `jax.tree_flatten` for this.

brax/jumpy.py
vuoristo: I think ideally that would be how this worked. Right now though, jumpy uses `onp` by default and only uses `jnp` when needed. What setting this flag to numpy should do in practice? 

Forcing `jnp` is needed because `vmap`, `scan`, etc cannot know what kinds of inputs the functions they return will be called on when jit is disabled. This use case is asymmetric, i.e., forcing the `jnp` branch in `_in_jit()` should always work (I think), but forcing the `onp` branch wouldn't work when the functions are called with jax inputs.

Forcing `jnp` cannot actually change what `_which_np()` (without arguments) returns, as explained in the comment below. Therefore, forcing `jnp` doesn't actually change jumpy to exclusively use `jnp`. So in this current form, this isn't really a flag for selecting which backed to use but rather a flag to circumvent the specific problem with `JAX_DISABLE_JIT`.

erikfrey: OK, thanks for thinking through these edge cases.  I agree with the the points you laid out here.

Last suggestion, then: Given your line of thinking, the flag name "JUMPY_FORCE_JNP" might be too broad, because it's not really forcing jnp across the board.  So that leads me to two questions/suggestions:

1) What if we dropped the flag entirely, and instead `_in_jit()` always returns true if it detects `JAX_DISABLE_JIT` has been set?  This might be a bit too magical/unadvertised behavior... there might be a valid reason to set `JAX_DISABLE_JIT` and still use numpy code... but that's doubtful.  What do you think?

2) If that doesn't make sense, let's make the flag name a bit more specific: `JUMPY_IGNORE_DISABLE_JIT` or something.

vuoristo: It's indeed hard to come up with a reasonable case where you would want to disable jit and use numpy. If this gets merged, it's almost certain someone will come up with one though 😄. Anyway, I think this might be adequate workaround for now. I believe the correct flag to check would then be `jax.config.jax_disable_jit`.

brax/jumpy.py
vuoristo: This is a separate issue, but I think it would be better if brax always used `onp` explicitly in cases where `jnp` would not work. With that change, you could make it possible to choose the jumpy backend.

erikfrey:
Thanks for discovering this bug and working out a fix!

PR #212, opened 2022-07-30
Updated deprecated jax.tree_map to jax.tree_util.tree_map

YigitDemirag:
I recently started getting `FutureWarning: jax.tree_leaves is deprecated,  and will be removed in a future release. Use jax.tree_util.tree_leaves instead.`.  Hence, I refactored the code to reflect this JAX update.


btaba:
Hi @YigitDemirag, thanks for the PR and thanks for using brax! We just pushed out a fix for https://github.com/google/brax/issues/213 which had an internal update to use `jax.tree_util.tree_leaves`. Sorry for not syncing this change earlier!
Please let us know if there are any issues and please don't hesitate to contribute in the future!

PR #214, opened 2022-08-08
Fixed broken hyperlinks in readme

Kaleb-Bruwer:
Not a major change, just some hyperlinks that were presumably broken by a refactor

PR #215, opened 2022-08-10
Fixed links to examples in readme

MischaPanch:
This is a tiny PR to fix broken links


MischaPanch:
I now noticed that there already is PR #214 which redirects to the module instead of the training script. Feel free to close mine in favor of that one (or vice versa)

erikfrey:
Thanks Mischa!  We merged the earlier PR so we'll close this one.

PR #216, opened 2022-08-11
Fix force action dim bug and joint dof bug

GengYiran:
Hello, I found two bugs and fixed them. I hope this will help brax.
1: When applying multiple forces to a body, the dimension of action only counts one of the forces (only +3), after my fix it will count each force (+3*n, n represents the number of forces).
2: I found that an environment must have an actuator, otherwise it will report an error, which doesn't make sense. I modified it so that no error is reported even when there is no actuator, and it can proceed normally.


erikfrey:
Hi @GengYiran - thanks for opening this PR.  We're well on our way to preparing [v2](https://github.com/google/brax/tree/main/brax/v2) for full release, which has a much more elegant handling of joint dofs and generalized forces vs actuators, so please do have a look there and see if it works for you.

I'm going to close this PR as it relates to v1 which is hopefully soon to be deprecated.

PR #234, opened 2022-10-09
fix: fix type hinting for `extra_loss_fns` in mappo

cemlyn007:
The type hinting for `extra_loss_fns` in `mappo.py` was using the old location of where `StepData` was located.

I noticed this when running the `notebooks/composer/composer.ipynb` because one of the cells would error.


erikfrey:
Thank you!

PR #235, opened 2022-10-09
fix: fix type hinting for `SimObserver`

cemlyn007:
Corrected the type hinting for the `SimObserver` `indices` argument.


erikfrey:
Thank you!

PR #236, opened 2022-10-09
fix: fix type hinting for `tasks.py`

cemlyn007:
Corrected the type hinting for the `race` `pos` argument in `tasks.py`.


erikfrey:
Thank you!

PR #237, opened 2022-10-09
fix: fixed `ExtraStepArgsWrapper`

cemlyn007:
Fixed the `step` method for the `ExtraStepArgsWrapper` defined in `braxlines`.


cemlyn007:
Or is [this](https://github.com/google/brax/blob/28660f4e74056d3f0dd5d06bfdee6fa42f0a4f1f/brax/experimental/braxlines/training/env.py#L64) wrong for assuming that the `core_env` has type `ComponentEnv` or `ExtraStepArgsWrapper` on the outer layer for example?

Is the intent of the `ExtraStepArgsWrapper` to skip all inner wrappers and only call the innermost environment step function or to just forward the `kwargs` to the innermost environment step function whilst still calling the step function of each intermediate wrapper?

erikfrey:
Yes, this is in fact a bug.  Thanks for this fix!

Issue #240, opened 2022-10-11
Wrong inertia and torque on Humanoid environment


imoneoi:
It seems that the inertia matrices of Humanoid environment are set arbitrarily (using identity matrices), and motor torques may be too large. I have strange behaviors in training like oscillating and flying around.

The bodies are identical as Gym MuJoCo Humanoid, and mass calculation are correct (assuming water density, 1000 kg / m^3).

BTW, what are the training parameters of caption video? The humanoid gait looks realistic.

imoneoi:
Update: I wrote code to infer the mass and inertia from geometry like MuJoCo does, and would like to PR this feature.

There is another question. Brax seems to use only the diagonal of inertia tensor. I think we can approximate by only taking the diagonal elements, or update the physics code to use a 3x3 tensor ?

cdfreeman-google:
Indeed, we typically use identity inertias in most of the environments, because the maximal-coordinate style methods are quite sensitive to the relative inertial scales between parts.  This has been qualitatively "good enough" for research on continuous control problems, because precise inertias only start becoming relevant once you want to do something like sim2real (and even then, you'd be surprised with how much you can get away with).

So, Bad News: this is working as intended, and we don't plan on fixing this in brax v1
Good News: brax v2 will be significantly more careful about this, and loads inertial data from mujoco xmls directly.  Additionally, the spring/pbd codepaths in v2 apply forces in the center-of-mass frame, which greatly simplifies the inertia math.

Long story short: thank you tremendously for looking into this, and I think you'll be much happier with how this is treated in v2 :D

erikfrey:
That said, big +1 to the quality of and care you showed in your PR.  I'd be overjoyed for you to contribute more in the future          :-)

imoneoi:
It's really sensitive. When I plugged in the correct inertia, the simulation become unstable (maybe I failed to tune the damping parameters). And using the identity inertia lead to different types of unrealistic behaviors.

Looking forward to brax v2. Are there any public available versions yet? I really need a fast JAX jittable physics simulation. I'm glad to contribute and improve it.

btaba:
Hi @imoneoi we posted a preview of brax v2 here:
https://github.com/google/brax/releases/tag/v0.1.0

where we implemented non-spherical inertias. We have a humanoid env here that can be used:

https://github.com/google/brax/blob/f0de0e56126b7cd7cdf941f0fc02f3480a02716a/brax/v2/envs/humanoid.py#L27

[...]

PR #241, opened 2022-10-12
Automatic infer inertia from geometry

imoneoi:
Automatic infer inertia from collider geometry like MuJoCo, replacing the arbitrarily set default values, as an identity matrix is often too large.

If mass / inertia is not set, it will be inferred from geometry. (assuming 1000 kg / m^3 density)

See https://github.com/google/brax/issues/240

### DIFF
--- a/brax/physics/base.py
+++ b/brax/physics/base.py
@@ -222,10 +222,6 @@ def validate_config(
   ]):
     config.frozen.all = True
   for b in config.bodies:
-    inertia = b.inertia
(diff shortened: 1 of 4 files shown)


erikfrey:
Thank you for this PR!

Brax will soon support loading directly from mujoco xml files, and as part of that change we'll automatically calculate inertia just like mujoco does.  I'm going to close this PR in favor of that upcoming change, which we'll be pushing out a preview of in the coming days.

PR #244, opened 2022-10-21
fix: correct `fori_loop`

cemlyn007:
Good day 👋

If `jumpy.fori_loop` is meant to be compatible with `jax.numpy.fori_loop` then the function signature type hinting is incorrect and will raise an exception when not inside a jitted function.


erikfrey:
Hmmm, guess we hadn't used `fori_loop` anywhere had we?  Thank you for spotting this!

PR #268, opened 2022-12-01
Update math.py

cyprienc:
Adding type hints to rotate, inv_rotate, ang_to_quat. Correcting return type hint of quat_rot_axis.


erikfrey:
Thanks!

PR #282, opened 2022-12-22
Added loading mass and joint limits in URDF model importer.

ManuCorrea:
I have added mass property in URDF and upper/lower revolute joint to Brax model importer.

Test updated accordingly.

I have not added the properties for inertia since the internal model config differs from the [official one](http://wiki.ros.org/urdf/Tutorials/Adding%20Physical%20and%20Collision%20Properties%20to%20a%20URDF%20Model), being the official one a 3x3 matrix.

Inertia property in Body is as a Vector3, I guess changing it to a matrix 3x3 can break other parts of the engine, if this change should be made I would gladly take it.

https://github.com/google/brax/blob/7eaa16b4bf446b117b538dbe9c9401f97cf4afa2/brax/physics/config.proto#L24-L35

<br/>

How would the other limit properties fit in the engine computations? I tried to look for effort and velocity.

I saw limit_strength but without being sure of the units it used I didn't want to break anything.

For the case of velocity I did not find any similar parameter in the back end, are joints speed limited?
<br/>
Regarding joint types "universal" and "spherical" it was not clear to me how are they written in URDF, they are not official. I saw similar ones in Gazebo, SDF and MuJoCo. Due to this I left them as they were.
<br/>

### DIFF
--- a/brax/tests/urdf_test.py
+++ b/brax/tests/urdf_test.py
@@ -76,7 +76,8 @@ class UrdfTest(absltest.TestCase):
     self.assertEqual(config.joints[0].name, 'test_joint')
     self.assertEqual(len(config.actuators), 1)
     self.assertEqual(config.actuators[0].name, 'test_joint')
-
+    self.assertEqual(config.bodies[0].mass, 1.0)
+    self.assertEqual(config.bodies[1].mass, 2.0)
 
 if __name__ == '__main__':
   absltest.main()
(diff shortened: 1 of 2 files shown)


btaba:
Hi @ManuCorrea ! Thanks for the PR, it looks good! FYI we'll be using mujoco's URDF loader going forward (see example here https://github.com/google/brax/blob/main/brax/v2/io/mjcf_test.py#L133-L136).

PR #295, opened 2023-01-19
Added rgba colour from MuJoCo file to renderer

ManuCorrea:
Removed hardcoded 'target' colour since it can be specified in XML files.

Also added dependencies in setup to run brax/v2/visualizer/visualizer.py

I am also working on adding textures to the renderer. Using the current way all the information is dumped into the brax system json. Encoding de picture in the json makes the file really big and can slow down the process. I can try to finish out the texture feature so you could judge and we could discuss the implementation if necessary,

----------------------------------
I was not sure how to run all the tests until I realized with pytest, that is why I commited without the test passing.

Since my decision seems that would required changes in the rest of the project, should be convenient a default rgba value in [Geometry class](https://github.com/google/brax/blob/673a41f780fe0f137507d1c35286577c48bde4d1/brax/v2/base.py#L30)?  With Optional argument all the parts in the rest of the code should be modified to add None. Which way should I proceed, any suggestion or I find out my way?


ManuCorrea:
I have seen that with Python 3.10 it can be set defaults with parent and children dataclasses without problems.

If I put rgba as default None it errors because its default before the arguments in the children Classes:
```
@struct.dataclass
class Geometry(_Base):
  link_idx: Optional[jp.ndarray]
  transform: Transform
  friction: jp.ndarray
  elasticity: jp.ndarray
  # TypeError: non-default argument 'radius' follows default argument
  rgba: Optional[jp.ndarray] = None
```

I could do something like this for each of the children Classes:
```
@struct.dataclass
class Sphere(Geometry):
  """A sphere.

  Attributes:
    radius: radius of the sphere
  """

  radius: jp.ndarray
  rgba: Optional[jp.ndarray] = None


@struct.dataclass
class Capsule(Geometry):
  """A capsule.

  Attributes:
    radius: radius of the capsule end
    length: distance between the two capsule end centroids
  """

  radius: jp.ndarray
  length: jp.ndarray
  rgba: Optional[jp.ndarray] = None
```

brax/v2/io/mjcf.py
```diff
@@ -393,6 +393,7 @@ def load_model(mj: mujoco.MjModel) -> System:
         'transform': Transform(pos=mj.geom_pos[i], rot=mj.geom_quat[i]),
         'friction': mj.geom_friction[i, 0],
         'elasticity': custom['elasticity'][i],
+        'rgba': mj.geom_rgba[i], # if there is no colour specified, default is applied
```
btaba, line 396: Thanks for doing this!
Can we make the default rgba="0.4 0.33 0.26 1", to match the current brax viz, only if a user default is not specified? You might have to read this from the xml itself to see if a default was specified by the user.
In other words: If a default is specified by the user, let's use that new default, otherwise let's use the brax default instead of the mjcf default of "0.5, 0.5, 0.5, 1."

brax/v2/base.py
```diff
@@ -335,12 +338,13 @@ class Box(Geometry):
...
 
   halfsize: jp.ndarray
+  rgba: Optional[jp.ndarray] = None
 
 
 @struct.dataclass
 class Plane(Geometry):
   """An infinite plane whose normal points at +z in its coordinate space."""
-
+  rgba: Optional[jp.ndarray] = None
```
btaba, line 347: I think we'll want a `= struct.field(pytree_node=False)` here

brax/v2/base.py
```diff
@@ -335,12 +338,13 @@ class Box(Geometry):
   """
 
   halfsize: jp.ndarray
+  rgba: Optional[jp.ndarray] = None
 
 
 @struct.dataclass
 class Plane(Geometry):
   """An infinite plane whose normal points at +z in its coordinate space."""
```
btaba, line 346: could you update the docstrings to something like, `rgba: (4,) the rgba to display in the renderer`

brax/v2/test_data/colour_objects.xml
```diff
@@ -0,0 +1,33 @@
+<mujoco model="ant">
+  <compiler angle="degree" coordinate="local" inertiafromgeom="true" />
+  <option timestep="0.01" collision="predefined" />
+  <custom>
+    <numeric data="0.0 0.0 0.55 1.0 0.0 0.0 0.0 0.0 1.0 0.0 -1.0 0.0 -1.0 0.0 1.0" name="init_qpos" />
```
btaba, line 5: this line throws an error now, you can just delete `custom`

brax/v2/test_data/colour_objects.xml
```diff
@@ -0,0 +1,33 @@
...
+  </asset>
+  <default>
+    <joint armature="1" damping="1" limited="true"/>
+    <geom conaffinity="0" condim="3" density="5.0" friction="1 0.5 0.5" margin="0.01" rgba="0.8 0.6 0.4 1"/>
+  </default>
+  <worldbody>
+    <geom conaffinity="1" condim="3" name="floor" pos="0 0 0" rgba="1 0 0.8 1" size="40 40 40" type="plane"/>
+    <body name="sphere" pos="0 0 0.25">
+      <joint axis="1 0 0" name="free1" pos="0 0 0" type="free"/>
+      <geom name="sphere" fromto="-0.5 0.0 0 0.5 0.0 0" pos="0 0 0" size="0.25" type="capsule"/>
```
btaba, line 18: for a sphere `<geom name="sphere" pos="0 0 1" size="0.25" type="sphere"/>`

brax/v2/test_data/colour_objects.xml
```diff
@@ -0,0 +1,33 @@
...
+    </body>
+    <body name="capsule" pos="0 0 0.25">
+      <joint axis="1 0 0" name="free2" pos="0 0 0" type="free"/>
+      <geom name="capsule" fromto="-0.5 0.0 0 0.5 0.0 0" pos="0 0 0" size="0.25" type="capsule"/>
+    </body>
+    <body name="box" pos="0 0 0.25">
+      <joint axis="1 0 0" name="free3" pos="0 0 0" type="free"/>
+      <geom name="box" fromto="-0.5 0.0 0 0.5 0.0 0" pos="0 0 0" size="0.25" type="capsule"/>
```
btaba, line 26: for a box `<geom name="box" pos="0 0 0" size="0.25 0.25 0.25" type="box"/>`

brax/v2/io/mjcf.py
ManuCorrea: You are welcome! I can check if I could make that happen. Soon I will apply and test all the corrections and apply the  default if proceeds!

brax/v2/test_data/colour_objects.xml
ManuCorrea: Thank you, good observation, for my test didn´t really mind so I didn´t pay enough attention to that

brax/v2/base.py
ManuCorrea: With that in some of the tests I get `TypeError: __init__() missing 1 required positional argument: 'rgba'` it was bit tricky to me to find the way of adding it and getting no errors.

ManuCorrea: Also the other tests in mjcf_test.py have `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` so I think  maybe  leaving it as None is the straight way to go unless the solution is undesirable.

brax/v2/io/mjcf.py
ManuCorrea: Hello!

There is some part in load_model() that would be (not final implementation, some  guidance of the results in the prints):
```
    # (default_rgba info got from XML)
    if default_rgba:
      print(f'default provided, using {mj.geom_rgba[i]}')
    else:
      # mujoco default colour will be applied by mujoco reader
      """
      edge case if [0.5 0.5 0.5 1.] is provided in a children geom it would get overwritten
      there is no way of knowing if the value of mj.geom_rgba[i] is provided by user or mujoco
      unless cheching XML here for each of the geoms
      """
      if (mj.geom_rgba[i] == [0.5, 0.5,  0.5, 1. ]).all():
        print(f'default provided, using {[0.4, 0.33, 0.26, 1]} instead {mj.geom_rgba[i]}')
```
Checking the edge case may be too much and I think checking XML and the MuJoCo at the same time may be some technical debt. Should I skip the edge case and just proceed? I would move on the texture integration once this PR gets close.

btaba: maybe just this would do for now without getting `default_rgba` from the xml:

`'rgba': [0.4, 0.33, 0.26, 1.0] if (mj.geom_rgba[i] == [0.5, 0.5,  0.5, 1. ]).all() else mj.geom_rgba[i]`

brax/v2/base.py
btaba: So the `TypeError: __init__() missing 1 required positional argument: 'rgba'` seems to be an error in barx.v2.geometry.mesh that needs resolving, at least on my branch. I'm not sure how you're getting the ValueError
Quite a bit changed in mjcf and base.System.geoms, and you're right that we don't want the pytree_node=False :)

brax/v2/io/mjcf.py
ManuCorrea: That will upset the users that really want [0.5, 0.5,  0.5, 1. ] but we can live with that! :laughing:

btaba: @ManuCorrea could you push your latest commits? seems like most resolved comments are not resolved in the code, but LGTM otherwise

ManuCorrea: Yes! Sorry I didn't want to push the partially resolved stuff until having all and tested, I have been busy recently.

ManuCorrea:
I have seem quite some has evolved in the repo since the PR! I will work on the texture implementation soon, should I also commit in the PR the work in progress for having a more continuous feedback?

btaba:
> I have seem quite some has evolved in the repo since the PR! I will work on the texture implementation soon, should I also commit in the PR the work in progress for having a more continuous feedback?

Hey @ManuCorrea , I'm not sure what you mean by the work in progress, but this PR looks good, let's get it in! Just some failing tests to resolve on this branch

btaba (changes requested):

ManuCorrea:
Now with the new mjcf.py version the tests fails. Seems to be that it generated Convex objects which has rgba value as None and the arguments from mjcf.py are not passed to the Convex object in base.py but to a constructor from mesh.py

PR #307, opened 2023-02-03
Writer: replace add_scalars with add_scalar

gijskoning:
Both functions show the logged values in Tensorboard, however 'add_scalar' is much cleaner: 
Previously add_scalars would create a new run folder for each tag, and the description would contain the key twice (e.g. key='eval/reward' results in description: run/eval/reward/**eval/reward**). This resulted in Tensorboard showing multiple runs while actually only doing a single run with different tags. 

The new method can be combined with the old method since both will still write to the same summaryboard.


erikfrey:
Good find, thank you!

Issue #297, opened 2023-01-23
`Transform.inv` seems incorrect


SimiPixel:
Dear Brax authors,

while scrolling through some of the source i stumpled upon

```python
class Transform(_Base):
  def inv(self):
    """Invert the transform."""
    return Transform(pos=-1.0 * self.pos, rot=math.quat_inv(self.rot))
```

Is this really correct? Is it due to some convention i am unaware of?

It also seems inconsistent with Roy Featherstone 
![Screenshot from 2023-01-23 14-30-09](https://user-images.githubusercontent.com/61733064/214051820-9f36b46b-1361-4cf9-9325-8be71bc69ee9.png)

Greetings

btaba:
Hi @SimiPixel we follow a similar convention to [TDS](https://github.com/erwincoumans/tiny-differentiable-simulator/blob/24c1d2a820e4e8a46d2036fe206340615085c8ef/src/math/transform.hpp#L13), where `Transform` is expressed as a translation and quaternion rotation, rather than a 6x6

PR #316, opened 2023-02-28
Fix implementation of Transform.inv()

allanzhao:
I've been porting some of my research code to Brax v2, and the performance is very good so far! Our code uses rigid body transforms extensively, and we noticed that `Transform.inv()` seems to differ from the textbook formula (as mentioned in #297). 

This pull request:
1) changes the implementation of `brax.v2.base.Transform.inv()` to match [tiny-differentiable-simulator](https://github.com/erwincoumans/tiny-differentiable-simulator/blob/24c1d2a820e4e8a46d2036fe206340615085c8ef/src/math/transform.hpp#L191) and [MuJoCo](https://github.com/deepmind/mujoco/blob/50bebcb4ee78a3f29644cd081e9f570baf9e4e06/src/engine/engine_util_spatial.c#L302)
2) adds a test case at the end of `brax/v2/math_test.py` (please feel free to move to an appropriate place)

### DIFF
--- a/brax/v2/base.py
+++ b/brax/v2/base.py
@@ -130,9 +130,10 @@ class Transform(Base):
     rot = math.quat_mul(math.quat_inv(t.rot), self.rot)
     return Transform(pos=pos, rot=rot)
 
-  def inv(self):
+  def inv(self) -> 'Transform':
     """Invert the transform."""
-    return Transform(pos=-1.0 * self.pos, rot=math.quat_inv(self.rot))
+    inv_rot = math.quat_inv(self.rot)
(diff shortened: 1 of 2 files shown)


btaba:
Hi @allanzhao ! Thanks for the PR and documenting the bug!
We removed Transform.inv() in favor of `inv_do` 
https://github.com/google/brax/blob/9d5a88ef69606966ad075a7b043b5d1541342991/brax/v2/base.py#L127-L129

the reason being that Transform.inv(), as implemented here, is wrt other Transforms, not necessarily wrt a Motion or a Force which is a bit confusing. We were only using Transform.inv_do(Motion), but if you want to change this into a Transform.inv_do(Transform) operator then that would be great!

PR #318, opened 2023-03-09
Fix the dependency on dataclasses

wang-r-j:
Only require dataclasses if python<3.7.

As of python 3.7, dataclasses is a built-in library. Newer versions of dataclasses require python<3.7 (ericvsmith/dataclasses@44724c4). If dataclasses is required in python>=3.7, the older version of dataclasses (with bugs) will be forced to install, which is useless and not expected.


erikfrey:
Thank you!

Issue #331, opened 2023-04-06
Visual Observations and video recordings?


StoneT2000:
Just wondering if this is possible at the moment, it seems the old brax.io.image module is now gone in brax v2.

Moreover, if there is rendering support, are they GPU parallelizable or is it just state now (with single thread rendering for an eval env)

Thanks!

Issue #108, opened 2021-11-10
TPU/GPU visualization for learning from pixels


erikfrey:
Brax's rendering is CPU-only at the moment.  For agents with vision, it would be useful for rendering to happen on accelerator so that training can stay fast.

[Vispy](https://vispy.org/) may be useful, but a fast native jax renderer would be even better for data on device compatability.

masud99r:
Thanks for the great simulator. Any update on TPU/GPU visualization to enable learning from pixels?

yardenas:
Hi!
I would be happy to try and help, any ideas where to start?
Could [this](https://blog.evjang.com/2019/11/jaxpt.html) be of any use?

erikfrey:
Hi @yardenas and @masud99r - JAX rendering is still an area of active interest for us!  One of those libraries might be a good starting point, although you'll also need to generate meshes from our brax primitives like capsules.

You can find examples of how to do that in our current cpu-only renderer code: https://github.com/erwincoumans/tinyrenderer

If you get on-device rendering going via JAX, please share with us a colab so we can check it out!

yardenas:
Thanks @erikfrey for the hints!

So just trying to get a better understanding: basically we would need to implement functions like ```create_capsule``` ([e.g. here](https://github.com/google/brax/blob/8a3ea6065866f687cf657d4fc5c91d002a75f72f/brax/io/image.py#L69)) which takes in the ```radius```, ```half_height``` etc, but have it written in JAX, right?

SimiPixel:
> Brax's rendering is CPU-only at the moment. For agents with vision, it would be useful for rendering to happen on accelerator so that training can stay fast.
> 
> [Vispy](https://vispy.org/) may be useful, but a fast native jax renderer would be even better for data on device compatability.

I am currently looking in this direction for my personal projects. I made a post there how one might do this efficiently. Might be interesting here too. [Issue over at VisPy](https://github.com/vispy/vispy/issues/2448)

JoeyTeng:
Hi, I have created a pure JAX implementation and an adapter layer mimicking the behaviour of existing [pytinyrenderer](https://github.com/erwincoumans/tinyrenderer). The code will be released under Apache-2.0 License [here](https://github.com/JoeyTeng/jaxrenderer/tree/v0.1.0). I will be working on a PR to replace current visualisation code in Brax soon.

[...]

Issue #67, opened 2021-10-09
From pixels?


slerman12:
Is there a way to get the states as image frames?

erikfrey:
Hi Sam,

We're pushing an update with support for rendering states on Tuesday or Wednesday.  Stay tuned!

erikfrey:
OK, rendering to pixels is now available, but not yet documented.  If you'd like to try it, please check out:

https://github.com/google/brax/blob/main/brax/io/image.py

In the coming days we'll update our colabs with an example of how to use it.

erikfrey:
Closing this as I think it's addressed, but please do re-open if you have any other questions.  We are tracking the request to render using accelerators at #108

Issue #47, opened 2021-08-31
Rendering without notebooks


namheegordonkim:
Hi all,

Is rendering without notebooks supported? I personally don't like using notebooks and am much more used to working with e.g. GLUT windows. It seems that wrapping brax environment into `gym` as per the third tutorial (https://colab.research.google.com/github/google/brax/blob/main/notebooks/pytorch.ipynb#scrollTo=9mqL2S67losC) doesn't let me call `env.render()` (throws `NotImplementedError`).

Any thoughts would be highly appreciated. Thanks!

kayuksel:
Yes, that would be a great feature, even if that means not rendering whole batch but a selected one.
If we could render all of them in a single scene (where they do not collide), that could also be good.

bayerj:
I have been using [streamlit](https://streamlit.io/) for showing the renderings. If you have the html, it is as easy as this:

```python
import streamlit.components.v1 as components
html_string = load_html_rendering()    # this would have to be dumped in a different process
components.html(html_string, height=500)
```

Then reloading the streamlit app shows it. That's also nice to show all kinds of other analytics. Here is a screenshot of an app I built:

<img width="581" alt="Screenshot 2021-09-07 at 21 25 17" src="https://user-images.githubusercontent.com/50352/132399968-a47a7631-94e4-4702-8438-6f746224fc6e.png">

erikfrey:
We have begun work on this.  No ETA yet but it's an active priority for us.

EelcoHoogendoorn:
Interesting; that streamlit thing looks pretty slick.

I was running into the same issues myself; not very familiar with 3JS and I think itd be cleaner to keep the visualization within python. Or at least have multiple rendering options available.

The best thing I can come up with would be to use the high level vispy/scene module; which is a similar high level API to 3JS; though its a bit stale in terms of maintenance... but I suppose its fine for putting some cylinders and boxes on screen. Not sure if there is a better option in the python ecosystem; doing it in raw pyopengl sure would be a kludge.

yardenas:
@erikfrey what's the timeline on the GPU rendering project? I might be able to contribute some parts.

PR #367, opened 2023-06-02
io/image: Replace PyTinyrenderer with a Pure JAX renderer

JoeyTeng:
This solves #331 #108 #47 #67 and is involved in #302

[Example Colab](https://colab.research.google.com/drive/14uV9ORm44lDNxmyjOvpOgGX-o-EVgkhS) (adopt majority of the code from Brax teams's [Brax Training](https://colab.research.google.com/github/google/brax/blob/main/notebooks/training.ipynb))

<s>Known issue: the plane may not display in full correctly, due to unclipped vertices behind/at the camera plane. This will be solved once a proper clipping algorithm is implemented in JaxRenderer (will be delivered soon).</s>
The plane rendering issue is now solved by implementing a rasterisation based on homogeneous interpolation.

Currently this only make changes in the v2 API, with minimal changes. Feel free to tell me if you think it is needed to back port to v1 API as well. I will try to optimise the performance further later.

Update: 
1. with 8a111d9, I jitted the rendering for each frame when renders a batch of states.
2. 828848a refactor batch rendering a bit but with no significant improvement in performance (if there is any), see [simple benchmark here](https://colab.research.google.com/drive/1gBIevFjnRrEpo2uU9blZ5qu6KIzDWTl7)
3. with JoeyTeng/jaxrenderer#2 ([release 0.3.0](https://github.com/JoeyTeng/jaxrenderer/releases/tag/v0.3.0)), the performance is improved to about 10x. Rendering one frame of the `Ant` environment with 960x540 resolution and 1x SSAA is now about [500ms using T4 (0.5 fps)](https://colab.research.google.com/drive/1xhkYNz5WjvUCjQWpp72CLf9SIy3i5PnN#scrollTo=hGLc2_6Kw4_L), and [~70-100ms using A100 (10-14 fps)](https://colab.research.google.com/drive/1A7PzhG3vn6oNzrWTxE5E3dmu8xQTcNnH#scrollTo=hGLc2_6Kw4_L). Previous CPU implementation is about [200ms per frame (5fps)](https://colab.research.google.com/drive/1xhkYNz5WjvUCjQWpp72CLf9SIy3i5PnN#scrollTo=SyVkqzz9oaJA).
4. with JoeyTeng/jaxrenderer#3 ([release 0.3.1](https://github.com/JoeyTeng/jaxrenderer/releases/tag/v0.3.1)), the minimum Python version is lowered to Python 3.8 which is the same as brax.

### DIFF
--- a/brax/io/image.py
+++ b/brax/io/image.py
@@ -15,7 +15,7 @@
 """Exports a system config and state as an image."""
 
 import io
-from typing import List, Optional, Tuple
+from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Sequence
 
 import brax
 from brax import base
@@ -27,8 +27,9 @@ from PIL import Image
 from renderer import CameraParameters as Camera
 from renderer import LightParameters as Light
 from renderer import Model as RendererMesh
+from renderer import ModelObject as Instance
 from renderer import ShadowParameters as Shadow
-from renderer import Renderer, Scene, UpAxis, transpose_for_display
+from renderer import Renderer, UpAxis, create_capsule, create_cube, transpose_for_display
 import trimesh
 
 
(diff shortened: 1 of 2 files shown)


JoeyTeng:
Friendly ping @erikfrey . With the release of jaxrenderer 0.3.0, I believe this should be a suitable replacement for current CPU renderer pytinyrenderer as 1) it is a pure JAX implementation; and 2) it performs better on high-end GPUs like A100 (>2x speedup). Let me know what you think :)

erikfrey:
Hi Joey - thanks for this tremendous effort!  Unfortunately we're bound by certain restrictions that make it very difficult for us to depend on external packages unless they go through an arduous vetting process behind the scenes.

I'm curious how your approach performs on CPU?

You've probably already found that using XLA on its own limits your performance here, as XLA wasn't designed for rasterization-like operations and it does not know how to use the underlying GPU primitives to actually do this performantly.

If you're interested, there is a way to register custom XLA ops, so that you could call into CUDA's rasterization functions.  Tensorflow graphics does such a thing:

https://github.com/tensorflow/graphics/blob/master/tensorflow_graphics/rendering/opengl/rasterizer_op.cc

You could then make such an op accessible to JAX.  This would be very fast, but also quite tricky to get all the plumbing to work.  Just thought I'd put it out there in case you're curious to hack on such a thing.

Either way though, it would be hard for us to accept this PR due to the policies that control which external packages we can rely on.  Sorry about that!  But I'll leave the issue open for a while - I'd love to hear if you continue hacking on this.

JoeyTeng:
Thank you so much for your suggestions! I am also thinking about customised op as that would dramatically improve the performance.

For the current performance, using high-end GPUs the rendering will actually be faster than CPU; current implementation is not optimised for TPU so the performance is really bad on TPUs, both for execution time and memory usage (up to 98% padding...). I will benchmark and improve the performance over a batch of small images (e.g. 84x84, as this is used in RL for Atori environment) to see if my implementation could benefit from rendering batches of environments in parallel (although we could definitely benefit if we are using a TPU Pod and pmap/xmap all environment simulation + rendering over devices).

sai-prasanna:
How about brax-contrib package or something like that where this can be added?

btaba:
We have since released demos with Madrona-MJX, I strongly recommend folks try that out for large-batch rendering on GPU.  https://github.com/shacklettbp/madrona_mjx

btaba:
Closing the PR since we recommend trying out https://github.com/shacklettbp/madrona_mjx as referenced here https://playground.mujoco.org/

btaba:
Never mind re-opening the PR as it is a great pedagogical example! But we don't expect to merge this in

Issue #338, opened 2023-04-20
Many randomized environments in parallel


jonasrothfuss:
I would like to use some form of domain randomization with brax environments. Ideally, I would like re-sample the parameters of the envs in every iteration / after every step. So far I only managed to sample random parameters and replace the default parameters in the environment XML. However, loading the 'system' from XML takes very long and makes this approach very inefficient. 

Is there a clean and efficient way to change relevant physics parameters (e.g. body mass, size, friction coef, etc.) without the significant overhead of reconstructing the 'system' from scratch?

btaba:
Hi @jonasrothfuss 

Yes there is a way to sample multiple systems, but we have not polished/released such a thing in v2. Tracing over the System is a high-level intent of the new design of brax. One way to get your project off the ground is to create a vectorized System like so:

```python
def _write_sys(sys, attr, val):
  """"Replaces attributes in sys with val."""
  if not attr:
    return sys
  if len(attr) == 2 and attr[0] == 'geoms':
    geoms = copy.deepcopy(sys.geoms)
    if not hasattr(val, '__iter__'):
      for i, g in enumerate(geoms):
        if not hasattr(g, attr[1]):
          continue
        geoms[i] = g.replace(**{attr[1]: val})
    else:
      sizes = [g.transform.pos.shape[0] for g in geoms]
      g_idx = 0
      for i, g in enumerate(geoms):
        if not hasattr(g, attr[1]):
          continue
        size = sizes[i]
        geoms[i] = g.replace(**{attr[1]: val[g_idx:g_idx + size].T})
        g_idx += size
    return sys.replace(geoms=geoms)
  if len(attr) == 1:
    return sys.replace(**{attr[0]: val})
  return sys.replace(**{attr[0]:
                        _write_sys(getattr(sys, attr[0]), attr[1:], val)})


def set_sys(sys, params: Dict[str, jp.ndarray]):
  """Sets params in the System."""
  for k in params.keys():
    sys = _write_sys(sys, k.split('.'), params[k])
  return sys

def randomize(sys, rng):
  return set_sys(sys, {'link.inertia.mass': sys.link.inertia.mass + jax.random.uniform(rng, shape=(sys.num_links(),))})

rng = jax.random.PRNGKey(0)
rng, *key = jax.random.split(rng, batch_size + 1)
key = jp.reshape(jp.stack(key), (batch_size, 2))
sys_v = jax.vmap(functools.partial(randomize, sys=sys))(key)
```

[...]

Issue #446, opened 2024-01-30
Subsume part of System inside State; EDIT: Or add Options to reset


joeryjoery:
For domain randomization it is not particularly easy to `vmap` over different `System` values. For example the `gravity` values, or the `elasticity`. Preferably you should be able to do this in `env.reset` but right now this is not possible as `self.sys` is a global variable in the Env namespace.

Right now my hacky workaround is to Mock the Brax environment with my custom PyTree-like dataclass so I can modify the `env.sys` values in a functionally pure way inside the `reset` function.

It would be nice if brax could expose part of the `sys` dict/ namespace as a pure argument to `env.reset` and `env.step` (e.g., as part of the state).

joeryjoery:
Wanted to add an example of another workaround: https://github.com/automl/CARL.

In this library for meta-RL, instead of batching environments on the GPU which Brax should support, the CARL-brax environments create VectorizedWrappers from Gymnasium in order to run multiple `System` variations simultaneously. Which kind of defeats the purpose of GPU parallelization....

btaba:
Hi @joeryjoery , I believe we considered passing around sys as part of the env state, but IIRC we managed to squeeze out better performance using the current implementation. 

https://github.com/google/brax/blob/a89322496dcb07ac5a7e002c2e1d287c8c64b7dd/brax/envs/wrappers/training.py#L199

Feel free to implement a version of the base env class and wrapper which passes the sys in a functional way (e.g. as part of the `state.info`).  If you manage to get the same training performance out of it, please send it our way!

joeryjoery:
> Hi @joeryjoery , I believe we considered passing around sys as part of the env state, but IIRC we managed to squeeze out better performance using the current implementation.
> 
> https://github.com/google/brax/blob/a89322496dcb07ac5a7e002c2e1d287c8c64b7dd/brax/envs/wrappers/training.py#L199
> 
> Feel free to implement a version of the base env class and wrapper which passes the sys in a functional way (e.g. as part of the `state.info`). If you manage to get the same training performance out of it, please send it our way!

Hey thanks for the reply. A big obstacle right now in trying to implement something like this is that the `pipeline.init` and `pipeline.step` functions are quite rigid. They only receive `self, q, qd, _debug` as arguments. 

[...]

PR #458, opened 2024-02-15
[FEATURE] Vsys feature: massively parallel domain randomization

Velythyl:
Hello!

For an unrelated research project, I needed a massively parallel RL environment with domain randomization capabilities. Isaac Sim/Gym/Omniverse fit the bill, but I also needed the simulator to be differentiable w.r.t. each domain randomization parameters.

So I set out to implement DR in brax. This is research code, so it's obviously a little janky and ad-hoc. But I thought maybe the brax community could find this interesting, and perhaps (with a lot of tuning) even merge it into brax main.

Special thanks to this github issue from which I stole some code ;)  [here](https://github.com/google/brax/issues/338)

Note that this domain randomization method is more powerful than [this](https://github.com/google/brax/blob/9e14acb7fd2697236eac7ec8df6d94579dab8123/brax/envs/wrappers/training.py#L199). With this code, we can randomize every single simulation step, if we so wish.

The summary of the implementation is simple: we just augment the simulation state to contain `sys`, thereby allowing every single parallel environment access to its own separate `sys`. Also, this enables us to resample `sys` according to some rule (for example, "resample every 50 steps").

---

# Features:

##### Domain randomization is controlled via a simple yaml file format that describes the path to a domain randomization target. Example: 

```
link:
  inertia:
    mass:
      base: [r, r, r, r, r, r, r]
      min: [-0.5, -0.5,-0.5,-0.5,-0.5,-0.5,-0.5]
      max: [0.5, 0.5,0.5,0.5,0.5,0.5,0.5]
  constraint_ang_damping:
    min: [-1,-1,-1,1,1,1,1]
    max: [2,2,2,1.5,1,1,1]
```

This randomizes over the 7 links of the robot. For the mass, the base is "r", so the value is "read" from the default value defined in the URDF file. The min-max ranges are both relative to the base, so the current setup randomizes from [r-0.5, r+0.5]. For the damping, no base is given, which defaults to "r". One could also set the base to a float value. Another possible value for the base is "n" ("none"), which disables randomization for this index.

##### Domain randomization is differentiable (!)

For example, running a simple optax optimizer, we can obtain the true domain randomization parameters in play for a specific timestep.

# Known issues:

- Because of the need for `sys` to be included in the state, a few of the python type hints are broken
- The yaml definition system is arbitrary and might not be best-practices
- This is kind of a huge PR, so thoroughly testing all ~800 lines of code changes is bound to be tough
- You can test the changes by looking at the script in the vsys wrapper's `if __name__ == "__main__":` function. Specifically, here: https://github.com/Velythyl/brax/blob/b6cab6449ba677108e37739286e0521f7c226a9e/brax/envs/wrappers/vsys.py#L553

---

Again, I don't expect this to be merged as-is. But perhaps the implementation might be interesting to the community, hence the reason for this PR.


lebrice:
Hey @Velythyl I've been looking forward to this feature for a while now, thanks a lot for sharing this!
I'm just curious, why did you close the PR?

Velythyl:
@lebrice Hey! Sorry, I realized I had some cleanup to do, and it was way past 5pm so I wanted to go home. I reopened it now.

brax/envs/half_cheetah.py
```diff
@@ -173,12 +173,12 @@ def reset(self, rng: jax.Array) -> State:
         'reward_ctrl': zero,
         'reward_run': zero,
     }
-    return State(pipeline_state, obs, reward, done, metrics)
+    return State(pipeline_state, obs, reward, done,sys,  metrics)
```
lebrice, line 176:
```suggestion
    return State(pipeline_state, obs, reward, done, sys, metrics)
```

brax/envs/hopper.py
```diff
@@ -218,12 +218,12 @@ def reset(self, rng: jax.Array) -> State:
         'x_position': zero,
         'x_velocity': zero,
     }
-    return State(pipeline_state, obs, reward, done, metrics)
+    return State(pipeline_state, obs, reward, done,sys,  metrics)
```
lebrice, line 221:
```suggestion
    return State(pipeline_state, obs, reward, done, sys, metrics)
```

lebrice (commented):
(I'm not a maintainer, this is just a fix for some spacing typos, this is very clean!)

btaba:
Thanks @Velythyl !  The recent comment made me just realize that maintainers hadn't commented on the PR. There were a few design decisions that went into [DomainRandomizationVmapWrapper](https://github.com/google/brax/blob/9e14acb7fd2697236eac7ec8df6d94579dab8123/brax/envs/wrappers/training.py#L199):

1. We saw better performance when `sys` was **not** added as part of  `State`
2. We wanted the user to fully define the randomization strategy rather than force a schema onto the user. At HEAD, this can be done via the [`randomization_fn`](https://github.com/google/brax/blob/9e14acb7fd2697236eac7ec8df6d94579dab8123/brax/envs/wrappers/training.py#L205).

The cons of the impl at HEAD are that:

1.  The reset is static and stored in the [wrapper](https://github.com/google/brax/blob/9e14acb7fd2697236eac7ec8df6d94579dab8123/brax/envs/wrappers/training.py#L208). 

What I think would make sense to merge, is to add a wrapper with the same API as `DomainRandomizationVmapWrapper`, that passes `in_axes` and the randomized Sys PyTree values in the `State`, as discussed in this thread:

https://github.com/google/brax/issues/446

btaba:
Closing due to lack of activity.

PR #476, opened 2024-04-15
Updated basic APG algorithm

Andrew-Luo1:
The goal of this proposed update is to provide a simple APG algorithm that can solve non-trivial tasks, as a first step for researchers and practitioners to explore Brax's differentiable simulation. It has been tested on MJX. Notes:
- A demonstration of this algorithm is shown [here](https://github.com/Andrew-Luo1/mujoco_apg_tutorial/blob/main/mjx/training_apg.ipynb). 
- I have not benchmarked this algorithm against Brax's RL algorithms such as PPO and SAC, since a) the environments would need differentiable rewards, and b) this simple implementation aims to be a basic reference to extend upon.

**1: Algorithm Update**

This fork contains an APG implementation that is about as simple as the current one, but reflects a common thread between several recent results that have used differentiable simulators to achieve locomotion: [1](https://arxiv.org/abs/2204.07137), [2](https://adaptive-horizon-actor-critic.github.io/), [3](https://arxiv.org/abs/2403.14864). 

Brax's current APG algorithm is roughly equivalent to the following pseudocode:
```
for i in range(n_epochs):
    reset state
    policy_grads = []
    for j in range(episode_length // short_horizon))
        state, policy_grad = jax.grad(unroll(state, policy, short_horizon))
        policy_grads.append(policy_grad)
    optimizer.step(mean(policy_grads))
```

In contrast, the cited results update the policy gradient much more frequently, using the observation that policy gradients that differentiate through the simulator have low variance. Hence, unrolling for an entire episode before updating has limited use. That additional samples past a certain point do not help is seen in that convergence does not increase with with massive parallelization [2]. The proposed APG algorithm essentially performs live stochastic gradient descent on the policy, unrolling it for a short window, doing a gradient update, then continuing where it left off:

```
reset state
for i in range(n_epochs):
    state, policy_grad = jax.grad(unroll(state, policy, short_horizon))
    optimizer.step(policy_grad)
```

Note that n_epochs can be much larger in this case. This modification allows the algorithm to relatively quickly learn quadruped locomotion, albeit with a particular training pipeline and reward design. ([Notebook](https://github.com/Andrew-Luo1/mujoco_apg_tutorial/blob/main/mjx/training_apg.ipynb))

Additional notes:
- This fork uses a learning schedule to improve training stability.
- The particular choices of Adam optimizer parameters come from [2] and significantly improve training outcomes.

**2: Supporting Updates**

*Configurable initial policy variance*: When hotstarting a policy, it benefits to explore around its induced state-space vicinity. This can be done by initializing the policy network weights small. Currently, the softplus disables this possibility, so this fork adds a scaling parameter.

*Layer norm*: I have found that using [layer normalization](https://arxiv.org/abs/1607.06450) in the policy neural network has greatly improved the training stability of APG methods and is seen in [other implementations](https://github.com/NVlabs/DiffRL/blob/a4c0dd1696d3c3b885ce85a3cb64370b580cb913/models/actor.py#L32).

### DIFF
--- a/brax/training/agents/apg/train_test.py
+++ b/brax/training/agents/apg/train_test.py
@@ -22,10 +22,6 @@ from brax.training.acme import running_statistics
 from brax.training.agents.apg import networks as apg_networks
 from brax.training.agents.apg import train as apg
 import jax
-from jax import config
-config.update("jax_enable_x64", True)
-config.update('jax_default_matmul_precision', jax.lax.Precision.HIGH)
-
 
 class APGTest(parameterized.TestCase):
   """Tests for APG module."""
--- a/brax/training/agents/apg/train_test.py
+++ b/brax/training/agents/apg/train_test.py
@@ -21,7 +21,9 @@ from brax import envs
 from brax.training.acme import running_statistics
 from brax.training.agents.apg import networks as apg_networks
 from brax.training.agents.apg import train as apg
-import jax
+from jax import config
+config.update("jax_enable_x64", True)
+config.update('jax_default_matmul_precision', jax.lax.Precision.HIGH)
 
 
 class APGTest(parameterized.TestCase):
@@ -31,6 +33,7 @@ class APGTest(parameterized.TestCase):
     """Test APG with a simple env."""
     _, _, metrics = apg.train(
         envs.get_environment('fast'),
+        policy_updates=200,
         episode_length=128,
         num_envs=64,
         num_evals=200,
@@ -45,13 +48,14 @@ class APGTest(parameterized.TestCase):
     env = envs.get_environment('fast')
     original_inference, params, _ = apg.train(
         envs.get_environment('fast'),
+        policy_updates=200,
         episode_length=100,
         action_repeat=4,
         num_envs=16,
         learning_rate=3e-3,
         normalize_observations=normalize_observations,
-        num_evals=200,
-        truncation_length=10)
+        num_evals=200
+    )
     normalize_fn = lambda x, y: x
     if normalize_observations:
       normalize_fn = running_statistics.normalize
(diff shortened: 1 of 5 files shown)


brax/training/distribution.py
```diff
@@ -131,7 +131,7 @@ def forward_log_det_jacobian(self, x):
 class NormalTanhDistribution(ParametricDistribution):
   """Normal distribution followed by tanh."""
 
-  def __init__(self, event_size, min_std=0.001):
+  def __init__(self, event_size, min_std=0.001, var_scale=1):
```
erikfrey, line 134: nit: add var_scale to Args section of docstring

erikfrey (changes requested):
Looks great, just one nit.

erikfrey:
Oh - it looks like you need to update [brax/training/agents/apg/train_test.py](https://github.com/google/brax/blob/main/brax/training/agents/apg/train_test.py) - changes should hopefully be minimal - please review the [failing test](https://github.com/google/brax/actions/runs/8695501482/job/23891040303?pr=476).

Also, please do sign the CLA.  Thank you!

Andrew-Luo1:
Hi @erikfrey, I have updated the tests and they pass on my local setup. I've also fixed the nit. I've signed the CLA, and the Checks tab is saying that my signing went through. Please let me know if there's anything missing.

brax/training/agents/apg/train_test.py
```diff
@@ -22,6 +22,9 @@
 from brax.training.agents.apg import networks as apg_networks
 from brax.training.agents.apg import train as apg
 import jax
+from jax import config
+config.update("jax_enable_x64", True)
```
erikfrey, line 26: these config changes are process-wide, so it's breaking other tests that expect default float width and precision.  we run all the tests in a single process.

the envs in this test are simple enough that hopefully they don't need these config changes.  can you try removing these config changes or otherwise tweak the tests so that `jax_enable_64` and the precision change are not needed?

erikfrey (changes requested):
Almost there!

Andrew-Luo1:
I removed the double precision toggle and the tests still run fine on my local setup. Let's see if this works :)

erikfrey:
Amazing, thank you!

Issue #421, opened 2023-11-10
Add Gymnasium Wrapper


vyeevani:
Lots of other frameworks seem to be switching to gymnasium. It'd be nice to also have a gymnasium wrapper in brax.

btaba:
Hi @vyeevani, is this what you're looking for or something else? 

https://github.com/google/brax/blob/16304037a36b1d9c8c0b3084f57d1159627b636b/brax/envs/wrappers/gym.py#L27

vyeevani:
It’s that except using Gymnasium instead of Gym. The imports would change and also the returns are different for the step API. For example, you’d need to also return truncations (that you can get from the EpisodeWrapper). On Nov 10, 2023, at 10:06 AM, btaba ***@***.***> wrote:﻿
Hi @vyeevani, is this what you're looking for or something else?
https://github.com/google/brax/blob/16304037a36b1d9c8c0b3084f57d1159627b636b/brax/envs/wrappers/gym.py#L27

vyeevani:
https://gymnasium.farama.org

vyeevani:
It's a simple enough lift. It seems like OpenAI Gym isn't going to be maintained anymore with the aim of switching everyone over to Gymnasium.

btaba:
Ok, yeah the APIs are subtly different. Feel free to open up a PR with `brax/envs/wrappers/gymnasium.py`

PR #486, opened 2024-05-14
Switch gym dependency to gymnasium

giotherobot:
Since `gym` is deprecated in favor of `gymnasium` (see https://github.com/openai/gym?tab=readme-ov-file#important-notice) here I propose the switch, allowing both installation as alternative to keep backwards compatibilty.

This allows for use of python>=3.10, that is incompatible with gym.

Possible fix for #421.

### DIFF
--- a/brax/envs/wrappers/gym.py
+++ b/brax/envs/wrappers/gym.py
@@ -17,9 +17,9 @@ from typing import ClassVar, Optional
 
 from brax.envs.base import PipelineEnv
 from brax.io import image
-import gym
-from gym import spaces
-from gym.vector import utils
+import gymnasium as gym
+from gymnasium import spaces
+from gymnasium.vector import utils
 import jax
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


lebrice:
Just a note: I think the Gym wrappers here use the gym step format, not the gymnasium step format with terminated and truncated. Perhaps these would also need to be updated, as well as the pytorch training notebook that used those wrappers.

btaba:
Hi @lebrice can you add gymnasium to the requirements, so that the builds pass? Thank you!

btaba:
Ah ok, sorry I was misreading this. I think the right thing to do is what @lebrice is saying. Switch to gymnasium altogether and hash out differences in the API calls in brax. You can ignore `brax/v1` entirely since it's deprecated. How does that sound?

btaba:
@giotherobot any updates? Will close the PR if there is no further activity

lupusorina:
Will this PR be merged in soon? Thanks.

Issue #535, opened 2024-10-15
Issue with rendering for VectorGymWrapper


bridgesign:
It seems that the code for the render function in `brax.envs.wrappers.gym.VectorGymWrapper` is wrong. The code is intended only to render the first environment and not all environments. In case of gym/gymansium, vector envs render all environments at the same time. The error comes on this [line](https://github.com/google/brax/blob/main/brax/envs/wrappers/gym.py#L152)

The issue is that `pipeline_state` does not implement `take`.

The code can be fixed by replacing the line with the following:
```python
  def render(self, mode='human'):
    if mode == 'rgb_array':
      sys, state = self._env.sys, self._state
      if state is None:
        raise RuntimeError('must call reset or step before rendering')
      # Change this line to return (env_num, height, width, 3)
      return np.stack((image.render_array(sys, state.take(i).pipeline_state, 256, 256) for i in range(self.num_envs))
    else:
      return super().render(mode=mode)  # just raise an exception
```
This will return an ndarray with images stacked for each environment.

PR #549, opened 2024-11-01
VectorEnv gym visulization correction

bridgesign:
This is a solution in context of https://github.com/google/brax/issues/535


bridgesign:
@btaba Is it possible to add any testing to ensure it does not break when changes are made?

btaba:
You can add a test in [`gym_test.py`](https://github.com/google/brax/blob/main/brax/envs/wrappers/gym_test.py).

bridgesign:
@btaba I have added tests. In addition, added kwargs in the method to allow width and height which is a part of the render function in gym api.

btaba:
@bridgesign thanks for the PR, can you fix the tests?

bridgesign:
@btaba I am not sure what the issue is. It works when I try locally. I think it is breaking because `render_array` is broken https://github.com/google/brax/pull/558.

btaba:
I don't think this is related to #558 . One of the tests you added is failing, maybe lower the batch size

bridgesign:
@btaba Done

bridgesign:
Just saw it failed again. I have absolutely no idea why the test is failing. 16 envs should not be too much right?

btaba:
@bridgesign maybe try a smaller image or just the non-batched test? Would be great if you can narrow it down

PR #559, opened 2024-11-26
Add dict obs support for PPO

Andrew-Luo1:
Implementation of @kevinzakka's specification for dictionary observations. Dictionary-valued observations are useful in a lot of contexts: [privileged critic inputs](https://arxiv.org/abs/1710.06542) and policy inputs that mix pixels and state information to name two.

**Behaviour**
1. In the case of ndarray-valued observations, this PR makes no changes.
2. When you call `ppo/train.py` with a dict-valued observation with observation normalisation enabled, it is applied to obs['state'].
3. When you call any other training agent with a dict-valued observation, it raises a NotImplementedError. Supporting other agents would be future work.

**Usage**
An upcoming PR on pixels-based PPO training provides an example of how to train with dictionary observations. Essentially, 
1. Ensure that `env.step` and `env.reset` return the observation in a dictionary form; see the example below.
4. Implement your environment's `observation_size -> Union[int, Mapping[str, Tuple[int, ...]]]` property. 
5. Change the called network, the `apply` method and the dummy observation generation in `make_policy_network` and `make_value_network`
6. Implement your desired network to process the dictionary observation.

*Example observation*
```python
obs = {
  'state': jnp.concat(data.qpos, data.qvel)
  'pixels/rgb': rgb/255.0
  'pixels/depth': depth/self._max_depth
}
```

**Side-effects**
`AutoResetWrapper` and `ppo/losses.py` needed to be slightly modified to support obs-dict training. These modifications do not appear to affect performance nor training (not shown), tested on the Franka Pick-up Cube task.

| Obs Type | Commit f43727 | PR     |
| -------- | ------------- | ------ |
| State    | 212118  SPS      | 213242 SPS |
| Dict     | NA            | 212729 SPS |


brax/training/agents/ppo/train.py
```diff
@@ -231,12 +231,17 @@ def train(
   key_envs = jnp.reshape(key_envs,
                          (local_devices_to_use, -1) + key_envs.shape[1:])
   env_state = reset_fn(key_envs)
+  ndarray_obs = isinstance(env_state.obs, jnp.ndarray) # Check whether observations are in dictionary form.
+  if not ndarray_obs and normalize_observations:
+    assert "state" in env.observation_size, "Observation normalisation only supported for states."
+
+  obs_shape = env_state.obs.shape[-1] if ndarray_obs else env.observation_size
```
btaba, line 238: Let's not rely on env.observation_size, I believe in many cases, that calls an env.step

brax/training/agents/ppo/train.py
```diff
@@ -322,10 +328,13 @@ def f(carry, unused_t):
     assert data.discount.shape[1:] == (unroll_length,)
 
     # Update normalization params and normalize observations.
-    normalizer_params = running_statistics.update(
-        training_state.normalizer_params,
-        data.observation,
-        pmap_axis_name=_PMAP_AXIS_NAME)
+    if normalize_observations:
```
btaba, line 331: why do we need this if statement, and the one that was added below it?

brax/training/types.py
```diff
@@ -79,7 +79,7 @@ class NetworkFactory(Protocol[NetworkType]):
 
   def __call__(
       self,
-      observation_size: int,
+      observation_size: Union[int, Mapping[str, Tuple[int, ...]]],
```
btaba, line 82: should this be Union[Tuple[int, ...], int] in the Mapping?

btaba (commented):
Thanks Andrew!

btaba:
@Andrew-Luo1 how do AutoResetWrapper and losses need to be modified? We should probably apply those updates here as well?

brax/training/types.py
Andrew-Luo1: Yes there seems to be no reason to enforce having <=1D obs be tuple-sized

brax/training/agents/ppo/train.py
Andrew-Luo1: switched to using jax.tree_util.tree_map(lambda x: x.shape, env_state.obs)

brax/training/types.py
Andrew-Luo1: In the case of a dict, observation_size will always be Mapping[str, Tuple[int, ...]] the way I calculate it using tree_map, but in general having non-tuple sizes makes sense.

brax/training/agents/ppo/train.py
Andrew-Luo1: For line 392, this handles the case when obs doesn't have the ['state'] key. Not sure if it makes sense for us to cover this case.

For the block under 331, there's nothing to update the normaliser with if 'state' isn't in the obs dict.

brax/training/types.py
```diff
@@ -79,7 +79,7 @@ class NetworkFactory(Protocol[NetworkType]):
 
   def __call__(
       self,
-      observation_size: int,
+      observation_size: Union[int, Mapping[str, Union[Tuple[int, ...], int]]],
```
btaba, line 82: May as well make this `ObservationSize` like the other types above

brax/training/agents/ppo/train.py
```diff
@@ -231,12 +231,16 @@ def train(
   key_envs = jnp.reshape(key_envs,
                          (local_devices_to_use, -1) + key_envs.shape[1:])
   env_state = reset_fn(key_envs)
+  ndarray_obs = isinstance(env_state.obs, jnp.ndarray) # Check whether observations are in dictionary form.
+
+  obs_shape = env_state.obs.shape[-1] if ndarray_obs \
+    else jax.tree_util.tree_map(lambda x: x.shape[2:], env_state.obs) # Discard batch axes.
```
btaba, line 237: Why do we have [2:] for the dict, buy [-1] for the jax.Array version? Presumably for pixel inputs?

Can they both be [2:] and then remove the if statement?

Can the comment be more explicit about why the first two dims are removed (one for num_envs, and one for num_devices, is that right?)

brax/training/agents/ppo/train.py
```diff
@@ -231,12 +231,16 @@ def train(
...
 
   normalize = lambda x, y: x
   if normalize_observations:
     normalize = running_statistics.normalize
   ppo_network = network_factory(
```
btaba, line 242: Don't we need to update the `network_factory` impl as well? Had an earlier comment along these lines

brax/training/agents/ppo/train.py
btaba: I think these if statements should be removed if the purpose is to validate whether "state" exists or not.

Let's fail with KeyError, as discussed previously, we don't want to fail silently in general

btaba (changes requested):

brax/envs/fast.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, **kwargs):
     self._dt = 0.02
     self._reset_count = 0
     self._step_count = 0
+    self._dict_obs = kwargs.get('dict_obs', False)
```
btaba, line 31: nit: `use_dict_obs`

`_dict_obs` makes it seem like you're storing dictionary observations to `self`

brax/envs/base.py
btaba, line 1: Should `def observation_size(` be updated in the `PipelineEnv`?

brax/training/networks.py
```diff
@@ -15,14 +15,15 @@
 """Network definitions."""
 
 import dataclasses
-from typing import Any, Callable, Sequence, Tuple
+from typing import Any, Callable, Sequence, Tuple, Mapping
```
btaba, line 18: nit: keep these imports in order, here and elsewhere in the PR

brax/training/networks.py
```diff
@@ -15,14 +15,15 @@
...
 import warnings
 
 from brax.training import types
 from brax.training.spectral_norm import SNDense
 from flax import linen
 import jax
 import jax.numpy as jnp
+from jax.tree_util import tree_flatten
```
btaba, line 26: nit: generally we try to avoid importing members of modules directly

brax/training/networks.py
```diff
@@ -82,48 +83,55 @@ def __call__(self, data: jnp.ndarray):
...
-    obs_size: int,
+    obs_size: types.ObservationSize,
     preprocess_observations_fn: types.PreprocessObservationFn = types
     .identity_observation_preprocessor,
     hidden_layer_sizes: Sequence[int] = (256, 256),
     activation: ActivationFn = linen.relu,
     kernel_init: Initializer = jax.nn.initializers.lecun_uniform(),
     layer_norm: bool = False) -> FeedForwardNetwork:
-  """Creates a policy network."""
+  """Creates a policy network. Only processes state in the case of dict obs."""
```
btaba, line 99: nit: this comment is better suited where it's applied rather than in the top-level function docstring. So on L107

Similar comment for value_network below

brax/training/agents/ppo/train.py
btaba: Thanks!

brax/training/networks.py
```diff
@@ -82,48 +83,55 @@ def __call__(self, data: jnp.ndarray):
         hidden = self.activation(hidden)
     return hidden
 
+def canolicalize_obs_size(obs_size: types.ObservationSize) -> int:
```
btaba, line 86: nit: "canonical" can be a quite overloaded term, let's call this `get_obs_state_size`

btaba (changes requested):

brax/envs/base.py
Andrew-Luo1: Will do

Andrew-Luo1: Please see the new version. It's a bit messy because observation_size can be an `int`, `tuple`, of a mapping of these. I propose to separate out the non-mapping int case so the other training algos don't break.

brax/envs/base.py
```diff
@@ -16,14 +16,15 @@
...
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 
 from brax import base
 from brax.generalized import pipeline as g_pipeline
 from brax.io import image
 from brax.mjx import pipeline as m_pipeline
 from brax.positional import pipeline as p_pipeline
 from brax.spring import pipeline as s_pipeline
+from brax.training.types import ObservationSize
```
btaba, line 27: We try to keep brax.training pretty standalone, can we remove this import?

brax/envs/base.py
```diff
@@ -139,10 +140,14 @@ def dt(self) -> jax.Array:
     return self.sys.opt.timestep * self._n_frames  # pytype: disable=attribute-error
 
   @property
-  def observation_size(self) -> int:
+  def observation_size(self) -> ObservationSize:
     rng = jax.random.PRNGKey(0)
     reset_state = self.unwrapped.reset(rng)
-    return reset_state.obs.shape[-1]
+    obs = reset_state.obs
+    if isinstance(obs, jax.Array) and len(obs.shape) == 1:
```
btaba, line 147: why do we need the len(obs.shape) == 1 ? just do what we had on the left hand side

brax/envs/base.py
```diff
@@ -139,10 +140,14 @@ def dt(self) -> jax.Array:
...
+      return obs.shape[-1]
+    else:
```
btaba, line 149: nit:

remove the else and just return

brax/training/networks.py
```diff
@@ -100,30 +103,36 @@ def make_policy_network(
       layer_norm=layer_norm)
 
   def apply(processor_params, policy_params, obs):
+    obs = obs if isinstance(obs, jnp.ndarray) \
```
btaba, line 106: nit: avoid backslash, use (...)

btaba (changes requested):
Looks good, just a couple more nits

brax/envs/base.py
Andrew-Luo1: Added clarifying comment. Does it make sense to have observation_size return a tuple for multi-dimensional obs?

btaba: Please fix

btaba (changes requested):

btaba (approved):

PR #560, opened 2024-11-26
PPO on Pixels

Andrew-Luo1:
**Logistics**
This PR is based on top of a [concurrent one](https://github.com/google/brax/pull/559) and can be rebased as needed. This PR's contributions are entirely in the four files:
```
brax/training/agents/ppo/cnn_networks.py
brax/training/agents/ppo/train_pixels.py
brax/training/agents/ppo/vision_networks.py
brax/training/agents/ppo/wrappers.py
```
**Results**
The PPO + networks setup introduced in this PR stably trains a pixel + prioprioceptive pick-up-cube task within 20 minutes and the pixel-based cartpole within a minute.

https://github.com/user-attachments/assets/980b601b-c4b8-4343-95b0-903558f27025

https://github.com/user-attachments/assets/30f8f6ab-d522-4a1d-bfa8-60b66f6a1685

**Algorithm**

*Usage*
Note that train_pixels.py *only* supports dictionary-valued observations. The pixel observations are expected to be under `obs['pixels/<name>']`, whereas any additional state-based observations are to be under `obs['state']`. 

*Image augmentation*
The key algorithmic difference between `ppo/train_pixels.py` and `ppo/train.py` is the option to randomly shift pixel data; toggled by the `augment_image` flag in `def train`. This implementation is a PPO adaptation of [DRQ](https://arxiv.org/abs/2004.13649) in the basic case of `K=M=1` (paper section 3.3). This simple change makes a significant improvement in the above franka pick-cube task, while not significantly affecting performance in cartpole balance.

Pasted image 20241122080755

**Networks**

This PR provides a `VisionMLP` - a basic network for pixel-based policies that combines a CNN encoder (the minimal NatureCNN architecture) and hidden layers afterwards. This network has three main features:
1. Support for multiple pixel streams, for example assigning a separate CNN to each camera views.
2. The option (`normalise_channels`) to independently apply layer norm to each independent pixel channel. This has been found to be useful when RGB and depth inputs are stacked for each camera view, since even when channels are normalised to [0, 1], depth inputs can be consistently larger in the case that most of the scene is background.
3. Appending any state-based observations to the CNN output encodings.

**Wrapping**

Under `ppo/wrappers.py`, this PR proposes a new version of the EpisodeWrapper that's optimised for vision.

The [*original* EpisodeWrapper ](https://github.com/google/brax/blob/main/brax/envs/wrappers/training.py#L93) scans over the entire env.step `action_repeat` times, generating `action_repeat - 1` unnecessary observations:

```python
    def f(state, _):
      nstate = self.env.step(state, action)
      return nstate, nstate.reward

    state, rewards = jax.lax.scan(f, state, (), self.action_repeat)
    state = state.replace(reward=jp.sum(rewards, axis=0))
```

In the case of pixel-based training, the additional overhead of unnecessary render calls becomes significant. 

Note that Brax offers a mostly equivalent mechanism of having each unroll see more simulated time without generating more RL transitions: the [under-the-hood scan of mjx.step](https://github.com/google/brax/blob/f43727eeebf21c031faf861ee00e98919c892140/brax/envs/base.py#L134).

If we call EpisodeWrapper's action repeat the `upper action repeat` (UAR), and the mjx.step scan the `lower action repeat` (LAR), `VisionEpisodeWrapper` unifies these two concepts by discarding the the UAR and assuming that you set the LAR to $LAR * UAR$ in the unwrapped environment. The episode wrapper then increments the steps counter by $LAR * UAR$ each step.

Let $l_e$ and $l_e'$ be the episode lengths for EpisodeWrapper and VisionEpisodeWrapper and $T$ and $T'$ similarly for the total training steps. To ensure the sample number of RL transitions and simulated time, set $l_e' = l_e * LAR$ and $T' = T * LAR$.

In practice, this new wrapping makes a significant performance improvment. On the above franka cube task, the raw sim throughput from looping `jit_step` goes from 99.8k to 173.9k simulation steps/second. The wall clock for the equivalent amount of simulated time and RL transitions goes from 25m:40s to 19m:28s.

**Hard-Coding**

Several aspects of this PR are specific to the upcoming pixels backend. In `train_pixels.py`, we remove the option to specify a separate `eval_env`, since the new backend does not allow multiple environment instances. For the same reason, we remove the `num_eval_envs` option.


btaba:
Thanks Andrew for the PR, will review it more thoroughly once the upstream PR is submitted.

One initial high level thought is that the wrapper should go in https://github.com/google/brax/blob/main/brax/envs/wrappers/training.py

brax/training/agents/ppo/train_pixels.py
```diff
@@ -0,0 +1,542 @@
...
+def _strip_weak_type(tree):
+  # brax user code is sometimes ambiguous about weak_type.  in order to
+  # avoid extra jit recompilations we strip all weak types from user input
+  def f(leaf):
+    leaf = jnp.asarray(leaf)
+    return leaf.astype(leaf.dtype)
+  return jax.tree_util.tree_map(f, tree)
+
+
+def train(
```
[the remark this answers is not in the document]
Andrew-Luo1: I was thinking that we might refine the pixel specific stuff more later on. But by that point, we might as well make a new folder. I’ve merged this into the main train.py.

Andrew-Luo1:
I've just rebased this PR on top of the [dict obs PR](https://github.com/google/brax/pull/559) and done the high-level refactoring just mentioned. It'd be great if you could take a look!

brax/training/types.py
```diff
@@ -30,7 +30,8 @@
 Params = Any
 PRNGKey = jnp.ndarray
 Metrics = Mapping[str, jnp.ndarray]
-Observation = jnp.ndarray
+Observation = Union[jnp.ndarray, Mapping[str, jnp.ndarray]]
```
btaba, line 33: can you please rebase to clean up the diff?

btaba (changes requested):

brax/training/agents/ppo/vision_networks.py
btaba, line 1: Let's maybe call these `networks_vision.py`, `networks_cnn.py`, `networks_*.py` etc.

brax/training/agents/ppo/cnn_networks.py
```diff
@@ -0,0 +1,91 @@
...
+ActivationFn = Callable[[jp.ndarray], jp.ndarray]
+Initializer = Callable[..., Any]
+
+
+class CNN(linen.Module):
+  """CNN module.
+  Warning: this expects the images to be 3D; convention NHWC
+  num_filters: the number of filters per layer
+  kernel_sizes: also per layer
+  """
```
btaba, line 24: nit: please consider running [black](https://black.readthedocs.io/en/stable/) on your PR

btaba (changes requested):

brax/training/agents/ppo/networks_vision.py
```diff
@@ -0,0 +1,149 @@
...
+    kernel_init=kernel_init,
+    layer_norm=layer_norm,
+    normalise_channels=normalise_channels,
+    state_obs_key=state_obs_key,
+  )
+
+  def apply(processor_params, policy_params, obs):
+    obs = FrozenDict(obs)
+    if state_obs_key:
+      state_obs = preprocess_observations_fn(remove_pixels(obs), processor_params)
```
erikfrey, line 66: i'm not quite sure i follow this.  it seems like you are doing this:

1) preprocess all non-pixel fields (i assume this is for normalization)
2) take only a SINGLE field (the state obs key) from the result and pass it into the module

if that's the case, why not just preprocess the state obs key?

brax/training/agents/ppo/train.py
```diff
@@ -218,14 +226,16 @@ def train(
           randomization_fn, rng=randomization_rng
       )
     if isinstance(environment, envs.Env):
-      wrap_for_training = envs.training.wrap
+      wrap_for_training = functools.partial(
```
erikfrey, line 229: ah i see

did you try instead changing the order of the wrapping in wrap_for_training such that VmapWrapper is created first?  Then you might not need this scan param

erikfrey (commented):
great start!

erikfrey (changes requested):
!

brax/envs/fast.py
```diff
@@ -19,34 +19,54 @@
...
 class Fast(PipelineEnv):
   """Trains an agent to go fast."""
 
-  def __init__(self, **kwargs):
+  def __init__(
+    self,
+    use_dict_obs: bool = False,
+    asymmetric_obs: bool = False,
+    pixel_obs: bool = False,
+    state_obs: bool = True,
```
btaba, line 33: these bools are getting a bit gnarly, can (use_dict_obs, pixel_obs and state_obs) be a single enum? From the user perspective, we could pass a string, and convert to enum as input validation within the class

btaba (changes requested):

erikfrey (approved):
Latest changes look great!  After you address @btaba 's last comment I'd say we are good to go.

brax/envs/fast.py
```diff
@@ -15,38 +15,68 @@
...
+
 import jax
 from jax import numpy as jp
+from flax.core import FrozenDict
+
+from brax import base
+from brax.envs.base import PipelineEnv, State
+
+
+class ObservationMode(Enum):
```
btaba, line 28: can you add a docstring? I had to scratch my head a bit to see what these modes meant

brax/training/agents/ppo/train.py
```diff
@@ -267,6 +283,36 @@ def train(
   gradient_update_fn = gradients.gradient_update_fn(
       loss_fn, optimizer, pmap_axis_name=_PMAP_AXIS_NAME, has_aux=True)
 
+  def random_translate_pixels(obs: Mapping[str, jax.Array], key):
```
btaba, line 286: nit: type annotation on key

also, does this function need to live inside `train()` or can it be moved outside?

btaba (changes requested):

brax/envs/wrappers/training.py
```diff
@@ -46,11 +46,11 @@ def wrap(
     environment did not already have batch dimensions, it is additional Vmap
     wrapped.
   """
-  env = EpisodeWrapper(env, episode_length, action_repeat)
   if randomization_fn is None:
     env = VmapWrapper(env)
   else:
     env = DomainRandomizationVmapWrapper(env, randomization_fn)
+  env = EpisodeWrapper(env, episode_length, action_repeat)
```
btaba, line 53: just OOC why is this move needed?

erikfrey: Something about how we implemented the batching rules that we use to call Madrona means that `vmap(scan(render()))` works but `scan(vmap(render()))` fails.

When we have some time we will fix this in `madrona_mjx` but for now it's easier to fix it here.

brax/training/agents/ppo/train.py
```diff
@@ -164,6 +175,11 @@ def train(
   Returns:
     Tuple of (make_policy function, network params, metrics)
   """
+  if madrona_backend:
+    assert not eval_env, "Madrona-MJX doesn't support multiple env instances"
```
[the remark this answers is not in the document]
erikfrey: or maybe even `raise ValueError(....`

brax/envs/wrappers/training.py
Andrew-Luo1: The Madrona Batch Renderer breaks when you do vmap(scan(env.step)) but is ok for scan(vmap(env.step)).

StafaH: This change is no longer needed, since Madrona requires it's own vmap wrapper to be used:
https://github.com/shacklettbp/madrona_mjx/blob/main/src/madrona_mjx/wrapper.py

wrap_env must be False when using vision, so that users know to use the MadronaWrapper

erikfrey: Nice.  I would be OK with merging this as is and then adopting the Madrona wrapper as a separate PR.

brax/training/agents/ppo/train.py
StafaH: It might be worth having an image_augmentations.py for future augmentations (color shifts etc.)

Andrew-Luo1:
Updated via feedback!

brax/training/agents/ppo/train.py
```diff
@@ -74,13 +74,46 @@ def f(leaf):
 
 
 def remove_pixels(obs: Union[jnp.ndarray, Mapping]) -> Union[jnp.ndarray, Mapping]:
```
btaba, line 76: last nit: let's keep these functions private since they are in `train.py` after all

btaba (changes requested):

brax/training/agents/ppo/train.py
Andrew-Luo1: Moved them :)

brax/training/agents/ppo/train.py
```diff
@@ -319,6 +278,45 @@ def train(
   gradient_update_fn = gradients.gradient_update_fn(
       loss_fn, optimizer, pmap_axis_name=_PMAP_AXIS_NAME, has_aux=True)
 
+  def random_translate_pixels(obs: Mapping[str, jax.Array], key: PRNGKey):
```
btaba, line 281: `_random_translate_pixels` etc. and should be moved back out for tidiness, sorry if that was unclear :|

btaba (approved):

erikfrey:
@Andrew-Luo1 can you follow the process to resolve conflicts, and then we will merge?

PR #567, opened 2024-12-06
Switch observations from FrozenDicts to plain Dicts

Andrew-Luo1:
[PR 560](https://github.com/google/brax/pull/560) assumed that enviroments would always return observations as `flax.core.FrozenDict`'s but this has already proven confusing/cumbersome, and it turns out that the Flax repository came to the [same conclusion](https://github.com/google/flax/discussions/3191).

This PR changes dictionary observation handling to assume them to be plain Python dictionaries.

### DIFF
--- a/brax/envs/fast.py
+++ b/brax/envs/fast.py
@@ -19,7 +19,6 @@ import enum
 
 from brax import base
 from brax.envs.base import PipelineEnv, State
-from flax import core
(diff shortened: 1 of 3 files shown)


btaba (approved):
Thanks for the cleanup @Andrew-Luo1 , looks good!

PR #582, opened 2025-03-10
Online Dagger for Behaviour Cloning

Andrew-Luo1:
This PR proposes an implementation of the Online Dagger algorithm, as described in [1]. This implementation supports both state and pixels-based workflows.

#### DAgger for Behaviour Cloning

Learning a policy directly from high-dimensional or noisy inputs—such as raw pixels—can be extremely challenging. **Behaviour Cloning (BC)**, also known as Policy Distillation or Teacher-Student Training, provides a supervised learning alternative to Reinforcement Learning (RL). This approach simplifies the control problem by dividing training into two phases. First, a robust teacher policy is developed using more informative inputs. Then, this teacher guides the training of a student policy through supervision signals. Compared to RL, BC often converges more efficiently and stably, and it eliminates the need for complex reward engineering.

A naive BC implementation might involve training the student solely on a static dataset of the teacher’s pre-recorded trajectories. However, this method is vulnerable to distribution drift: even a small mistake by the student can lead it away from the familiar distribution of the teacher's data, causing performance to deteriorate.

**DAgger (Dataset Aggregation)** offers an elegant solution to this problem at the cost of increased training time, due to additional data collection. Instead of relying solely on pre-collected teacher trajectories, DAgger continuously gathers data by allowing the student to generate its own trajectories during training. For each state in the student's trajectory, what the teacher would have done is used for the supervised learning signal, enabling the student to learn error recovery skills. This iterative loop helps maintain the student’s performance within the desired trajectory distribution. Training a teacher policy using RL and then transferring its expertise to a student via DAgger has become bread and butter in robotics research [2].

#### Outline

```
ALGORITHM: DAgger (Dataset Aggregation)
INPUT: Environment env, Teacher policy π_teacher, Initial student policy π_student
OUTPUT: Trained student policy π_student

FUNCTION Train(env, π_teacher, hyperparameters):
    Initialize training state with student policy π_student
    Initialize empty dataset D
    
    FOR i = 1 TO dagger_iterations:
        # Data Collection Phase
        β = dagger_beta_fn(i)  // Typically decreasing with iterations
        
        FOR each environment step:
            WITH PROBABILITY β:
                a_t = π_teacher(s_t)  // Query teacher for action
            ELSE:
                a_t = π_student(s_t)  // Use student's action
            
            Execute a_t in environment
            Record (s_t, π_teacher(s_t)) in dataset D  // Always label with teacher

        # Training Phase
        FOR epoch = 1 TO epochs:
            FOR each batch in D:
                Update π_student parameters via supervised learning
                (minimize distance between π_student(s) and π_teacher(s))
    
    RETURN π_student, metrics
```

#### Features
1. `num_envs`: Environment parallelization for faster and stabler training
2. `augment_pixels`: Random crop pixel augmentation. While this has not been seen to make a difference for training performance, interestingly it appears to be very important for sim2real of behaviour-cloned policies [3, 4]
3. `scramble_time`: The option to stagger the initial env timesteps, so every gradient update gets data from envs at different parts of the trajectory
4. `demo_length`: Compared to the original Dagger paper [5], Online Dagger, as the name suggests, has a much tighter loop between collecting data and updating the policy. By default, 6 steps are taken on the batched environments per gradient update. Perhaps surprisingly, this formulation trains stably

#### Sim to Real
This algorithm was used to train a pixels-based biarm manipulation policy that demonstrated [real-world transfer](https://github.com/Andrew-Luo1/mujoco_playground_new/blob/main/mujoco_playground/_src/manipulation/aloha/s2r/README.md). Skipping expensive evaluation runs, convergence requires about 2.5e6 simulator steps + renders - around 3 minutes wall-clock on a RTX4090.

#### Usage
For usage examples, please see train_test.py and train_dagger.py from the above link.

#### Missing Features
These can be added in the future if the community would find them useful.
1. *Multi-GPU Support*: Since behaviour cloning is so sample-efficient, supporting `jax.pmap` seems to have limited use-cases.
2. *Build-in Env Wrapping*: To reduce lines of code, I've left it to the user to apply env wrappers externally.
3. *Vector Obs Support*: This implementation assumes the env returns dictionary observations, as teacher and student networks often have access to different information.

#### References
[1] Lum, Tyler Ga Wei, Martin Matak, Viktor Makoviychuk, Ankur Handa, Arthur Allshire, Tucker Hermans, Nathan D. Ratliff, and Karl Van Wyk. “DextrAH-G: Pixels-to-Action Dexterous Arm-Hand Grasping with Geometric Fabrics.” arXiv, October 15, 2024. https://doi.org/10.48550/arXiv.2407.02274.
[2] Cheng, Xuxin, Kexin Shi, Ananye Agarwal, and Deepak Pathak. “Extreme Parkour with Legged Robots.” arXiv, September 25, 2023. https://doi.org/10.48550/arXiv.2309.14341.
[3] Mandlekar, Ajay, Danfei Xu, Josiah Wong, Soroush Nasiriany, Chen Wang, Rohun Kulkarni, Li Fei-Fei, Silvio Savarese, Yuke Zhu, and Roberto Martín-Martín. “What Matters in Learning from Offline Human Demonstrations for Robot Manipulation.” arXiv, September 25, 2021. https://doi.org/10.48550/arXiv.2108.03298.
[4] Chi, Cheng, Zhenjia Xu, Siyuan Feng, Eric Cousineau, Yilun Du, Benjamin Burchfiel, Russ Tedrake, and Shuran Song. “Diffusion Policy: Visuomotor Policy Learning via Action Diffusion.” arXiv, March 14, 2024. https://doi.org/10.48550/arXiv.2303.04137.
[5] Ross, Stephane, Geoffrey J. Gordon, and J. Andrew Bagnell. “A Reduction of Imitation Learning and Structured Prediction to No-Regret Online Learning.” arXiv, March 16, 2011. https://doi.org/10.48550/arXiv.1011.0686.


brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+from brax.training import gradients
+from brax.training import types
+from brax.training.acme import running_statistics
+from brax.training.acme import specs
+from brax.training.agents.bc import losses as bc_losses
+from brax.training.agents.bc import networks as bc_networks
+from brax.training.agents.ppo import train as ppo_train
+from brax.training.types import Params
+from brax.training.types import PRNGKey
+from brax.v1 import envs as envs_v1
```
btaba, line 39: delete v1 stuff please

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+  """ 
+  Optimisations:
+  1) No scanning over trajectory collection. Just have a lot of parallel envs.
+  2) Data never leaves the GPU. Only parameters flow in and out of it.
+  """
+
+  optimizer_state: optax.OptState
+  params: Params
+  normalizer_params: running_statistics.RunningStatisticsState
+  dagger_it: jp.ndarray
```
btaba, line 55: `dagger_step`? step is brax-y

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+    batch_size: int = 256,
+    scramble_time: int = 0,  # Maximum time to scramble the envs by.
+    network_factory: types.NetworkFactory[
+        bc_networks.BCNetworks
+    ] = bc_networks.make_bc_networks,
+    progress_fn: Callable[[int, Metrics], None] = lambda *args: None,
+    madrona_backend: bool = False,
+    seed: int = 0,
+    learning_rate=4e-4,
+    dagger_iterations: int = 1,
```
btaba, line 82: super nit-picky, dagger_steps

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+          init_params
+      ),  # pytype: disable=wrong-arg-types  # numpy-scalars
+      dagger_it=jp.array(0, dtype=int),
+  )
+
+  if (
+      restore_checkpoint_path is not None
+      and epath.Path(restore_checkpoint_path).exists()
+  ):
+    logging.info('restoring from checkpoint %s', restore_checkpoint_path)
```
btaba, line 160: we have a new checkpointing pattern, would be good to switch to that (see PPO agent)

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+    )
+    if '_steps' in state.info:  # Some envs have internal clocks.
+      state.info['_steps'] = jp.array(state.info['steps'])
+    return state
+
+  # Init the env state early to use for obs size calculations.
+  key, key_reset, key_scramble = jax.random.split(key, 3)
+  reset_fn = jax.jit(env.reset)
+  env_state = reset_fn(jax.random.split(key_reset, num_envs))
+  env_state = scramble_times(key_scramble, env_state)
```
btaba, line 123: is the idea here that the first episode will terminate potentially early, and subsequent episodes that contain successes will be staggered?

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+        lambda x: x.reshape((-1,) + x.shape[2:]), tree
+    )
+
+  assert (
+      int(num_envs * demo_length) % batch_size == 0
+  ), 'Demo trajectories must be evenly divisible into batches'
+  num_minibatches = int(num_envs * demo_length / batch_size)
+
+  @jax.vmap
+  def random_crop(key, img):
```
btaba, line 245: can this be pulled out as a private function up top?

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+    return env_state, obs, reward  # E x T
+
+  def collect_data(
+      env_state: envs.State,
+      key: PRNGKey,
+      params: Tuple[Params, Params],
+      beta: float,
+  ) -> types.Observation:
+    """Collect a dataset for behavioural cloning.
+    Builds out the dataset on host-side, assuming RAM > VRAM
```
btaba, line 215: wdym here that the dataset is on the host? Looks like it's on the device to me

brax/training/agents/bc/train.py
```diff
@@ -0,0 +1,399 @@
...
+    evaluate(training_state, {})
+
+  wall_time = 0
+  key, key_dagger = jax.random.split(key)
+  for i_eval in range(num_evals_after_init):
+    if i_eval != 0 and reset:
+      # Seed new randomness to be gradually phased in.
+      key_reset, key = jax.random.split(key)
+      _state = reset_fn(jax.random.split(key_reset, num_envs))
+      env_state.info['first_obs'] = _state.obs
```
btaba, line 377: shouldn't the reset handle this? Not sure why we should have anything playground specific

brax/training/networks.py
```diff
@@ -37,6 +37,42 @@ class FeedForwardNetwork:
...
+  bias: bool = True
+  layer_norm: bool = False
+  state_key: str = 'proprio'
+
+  @linen.compact
+  def __call__(self, data: jnp.ndarray):
+    latents = []
+    # Find all keys that match the pattern pixels/latent_{i} and sort them
+    latent_keys = sorted(
+        [k for k in data.keys() if k.startswith('pixels/latent_')],
```
btaba, line 58: can pixels/latent_ be parametrized? Where is this naming coming from?

brax/training/networks.py
```diff
@@ -37,6 +37,42 @@ class FeedForwardNetwork:
...
+        key=lambda x: int(x.split('_')[-1]),
+    )
+    for key in latent_keys:
+      latents.append(data[key])
+    hidden = [self.activation(linen.Dense(64)(latent)) for latent in latents]
```
btaba, line 63: should 64 be pulled out into an attribute?

brax/training/networks.py
```diff
@@ -37,6 +37,42 @@ class FeedForwardNetwork:
   apply: Callable[..., Any]
 
 
+class MLPLatents(linen.Module):
```
btaba, line 40: Would MLPHead be a better name for this?

btaba (changes requested):
Thanks Andrew!! This is a great addition to brax. I just have some small nits that should be quick fixes, let's get this in!

brax/training/agents/bc/train.py
Andrew-Luo1: Yep the student locks onto the teacher trajectory remarkably stably and teachers don't tend to terminate during the episode. So without this staggering, you get periodic loss spikes.

brax/training/agents/bc/train.py
Andrew-Luo1: This method of resetting preserves curriculum and other ideally persistent flags stored in info. However this seems to be less important for BC, so I think letting reset handle it makes sense.

brax/training/networks.py
Andrew-Luo1: Just for consistency with Aloha s2r code

Andrew-Luo1: Changing default to latent_

brax/training/networks.py
Andrew-Luo1: True MLPLatents is ambiguous on whether the latents are the inputs or outputs

brax/training/agents/bc/train.py
btaba: The env wrapper reset logic should be doing this automagically, if not please file a bug

btaba (approved):

Issue #49, opened 2021-09-01
Replacing gym's Mujoco envs with brax envs


vwxyzjn:
Had a conversation with @jkterry1 on https://github.com/openai/gym/issues/2366, and it appears brax would also be a great alternative for the mujoco envs replacement. 

To help with this transition. I made an attempt to try out brax with pytorch. Here is a basic report: https://wandb.ai/costa-huang/brax/reports/Brax-as-Pybullet-replacement--Vmlldzo5ODI4MDk. The source code is here: https://github.com/vwxyzjn/cleanrl/blob/mybranch/cleanrl/brax/readme.md

One of the biggest issue with the brax adoption is the env normalization:
* gym doesn't have a normalization wrapper
* sb3 has a normalization wrapper but brax does not have a sb3 vector env api
* brax's normalization is implemented in the training side (https://github.com/google/brax/blob/main/brax/training/normalization.py)

I think going forward, probably the best way to fix this is to refactor the brax training side's normalization to the environment side. This in the future will also help throughput with the `JaxToTorchWrapper`. Otherwise, the observation will go from GPU to CPU for gym or sb3's normalization wrapper, then GPU again for torch, which just doesn't make sense. 

One small thing is that given the brax environment directly produces the vector env, there is also no way to inject a `ClipActionsWrapper(env)`, which may or may not have a performance impact. That said, this can be implemented in the training side with ease.

erwincoumans:
Yes, as I suggested previously, Brax seems a good option, since it allows for GPU and TPU accelerators, next to CPU. We can use this issue to track progress and add an itemized todo.

jkterry1:
To recap the to do list:

1. Add suitable rendering
2. Further tune observation/action spaces to make them as close as possible
3. Make sure we are not reproducing the list of bugs in MuJuCo environments from Antonin Raffin that I sent you

I feel like there may have been a 4th issue, but I don't sleep very much and can no longer recall it. @erwincoumans @benelot do you remember?

vwxyzjn:
One note on the suitable rendering is I feel implementing `env.render(“rgb_array”)` might be too expensive and counterproductive. Maybe implementing `env.render(“html”)` at the end of training is more preferable.

jkterry1:
They're planning to add a new rendering engine such that "rgb_array" will be suitable

jkterry1:
[...]

PR #586, opened 2025-03-24
Add Normalization and Action Clipping Wrappers

cruz-lucas:
I implemented new wrappers designed for on-device normalization and action clipping to address the CPU–GPU transfer issues when using gym’s normalization wrappers. This contribution addresses [Issue #49](https://github.com/google/brax/issues/49). In particular, I added:

- **RunningMeanStd:** A class to compute running mean and variance for normalization.
- **ClipVecAction:** A wrapper that clips continuous actions to a specified range.
- **NormalizeVecObservation:** A wrapper that normalizes observations using running statistics computed by RunningMeanStd.
- **NormalizeVecReward:** A wrapper that normalizes rewards based on the running statistics of the accumulated (discounted) rewards.

I also added sanity check tests to verify the correct behavior of these wrappers. Please let me know if this doesn't align with Brax's design philosophy.

### DIFF
--- a/brax/envs/wrappers/training.py
+++ b/brax/envs/wrappers/training.py
@@ -250,3 +250,100 @@ class DomainRandomizationVmapWrapper(Wrapper):
         self._sys_v, state, action
     )
     return res
+
+
+class RunningMeanStd:
+    """Computes and updates running mean and variance for online normalization."""
+
+    def __init__(self, shape: Tuple[int], epsilon: float = 1e-8):
+        self.mean = jp.zeros(shape)
+        self.var = jp.ones(shape)
+        self.count = 0.0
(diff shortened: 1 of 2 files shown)


btaba:
Hi @cruz-lucas , is the running [statistic](https://github.com/google/brax/blob/241f9bc5bbd003f9cfc9ded7613388e2fe125af6/brax/training/agents/ppo/train.py#L533) pattern in agents not sufficient for your use-case?  Advantages can also be normalized in the PPO agent. ClipVecAction is not implemented, but it's usually simple enough to include the one-line jp.clip in the environment itself. What are your thoughts here?

btaba:
Closing the PR, but will re-open once we get more information about the change

PR #608, opened 2025-05-10
Fix type annotations

GameRoMan:


btaba:
Hi @GameRoMan thanks for the PR but we are no longer maintaining brax v1. We will be deleting the code in an upcoming PR.

PR #611, opened 2025-05-17
Update DomainRandomizationVmapWrapper (google#609)

AndruGomes13:
This change ensures that the original system (env.sys) remains unchanged
 during JAX tracing. This prevents the UnexpectedTracerError that would
 occur when wrapping the same environment twice.

Tests:
- Training on a local personal env works.
- Google Colab tutorial runs without needing the duplicate eval_env workaround.

### DIFF
--- a/brax/envs/wrappers/training.py
+++ b/brax/envs/wrappers/training.py
@@ -15,7 +15,8 @@
 # pylint:disable=g-multiple-import, g-importing-member
 """Wrappers to support Brax training."""
 
-from typing import Callable, Dict, Optional, Tuple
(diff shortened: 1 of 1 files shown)


AndruGomes13:
Hi everyone, I'm new to contributing here so I'm open to feedback!

Sorry for the PR spam, I’ve tried signing the CLA multiple times and reopening the PR, but it still shows as not signed. 

Thanks!

AndruGomes13:
Hi @btaba, could you provide some insights into the "import/copybara" failing check?

PR #660, opened 2026-02-26
Support callable entropy schedules in PPO

PhysicistJohn:
## Summary

PPO currently accepts a fixed entropy coefficient. Allow `entropy_cost` to also accept an Optax-compatible callable, supporting linear, cosine, warmup, piecewise, and custom schedules without new trainer flags. The scalar default and existing positional arguments are unchanged.

For example:

```python
entropy_cost = optax.cosine_decay_schedule(1e-2, decay_steps=1_000_000)
```

Pass this as `entropy_cost` to `ppo.train`. The schedule receives the total environment-step count, including `action_repeat`, once per rollout batch. All gradient updates on that batch share the coefficient. Counts start at zero for each train call, including parameter restoration. `training/entropy_cost` reports the mean applied coefficient.

## Testing

- 60 tests pass across PPO training, losses, checkpointing, and UInt64 types on CPU with Python 3.12 and JAX 0.11.1. This includes 11 new tests for schedule values, UInt64 rollover, scalar equivalence, float64 precision, and the environment-step clock.
- Four tiny PPO runs match upstream `bb77b328` exactly for default and explicit scalar coefficients with JAX x64 disabled and enabled, including parameters and common non-timing metrics.

### DIFF
--- a/brax/training/agents/ppo/losses.py
+++ b/brax/training/agents/ppo/losses.py
@@ -213,6 +213,7 @@ def compute_ppo_loss(
       'policy_loss': policy_loss,
       'v_loss': v_loss,
       'entropy_loss': entropy_loss,
+      'entropy_cost': entropy_cost,
       'kl_mean': kl,
       'policy_dist_mean_std': policy_dist_mean_std,
   }
--- a/brax/training/agents/ppo/train.py
+++ b/brax/training/agents/ppo/train.py
@@ -212,6 +212,8 @@ def train(
     # ppo params
     learning_rate: float = 1e-4,
     entropy_cost: float = 1e-4,
+    entropy_cost_end: Optional[float] = None,
+    entropy_schedule: str = 'linear',
     discounting: float = 0.9,
     unroll_length: int = 10,
     batch_size: int = 32,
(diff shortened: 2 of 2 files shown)


aaprasad:
@PhysicistJohn and brax team, whats the status of this PR, would be a really useful feature!

PhysicistJohn:
Hey, thanks for the ping. CLA is signed and passing now, so that part's
done. It's stuck behind Google's internal import/copybara review (the
"waiting for internal safe review approval" check above), that's on
their side, not something either of us can push from outside. No ETA
I can give you honestly. Will post here if I hear anything.

PhysicistJohn:
Updated this to accept an Optax-compatible callable through the existing `entropy_cost` argument, replacing the separate end-value and schedule-name flags. This supports cosine, linear, warmup, piecewise, and custom schedules while preserving existing scalar callers. I also merged current main. All 60 tests across PPO training, losses, checkpointing, and UInt64 types pass locally on CPU, including 11 new schedule regressions. Four tiny scalar-training comparisons match upstream exactly with JAX x64 disabled and enabled.

